PR #1, opened 2017-12-13
Changed tab bar navigator.

arcelena:


mobile/src/layout/tab-navigation-options-factory.tsx
```diff
@@ -1,19 +1,19 @@
 import React from 'react';
 import { NavigationTabScreenOptions } from 'react-navigation';
 import Icon from 'react-native-vector-icons/Ionicons';
+import { Image } from 'react-native';
+import { ImageURISource } from 'react-native';
 
 
 export class TabNavigationOptionsFactory {
-    public create(label: string, icon: string, unfocusedIcon: string): NavigationTabScreenOptions {
+    public create(label: string, focusedPath: ImageURISource, unfocusedPath: ImageURISource): NavigationTabScreenOptions {
```
santee: is it really an instance of `ImageURISource`? What does this thing look like in the console? I do not know the answer, just asking ;)

mobile/src/layout/root-navigator.tsx
```diff
@@ -9,26 +9,59 @@ import { TabNavigationOptionsFactory } from './tab-navigation-options-factory';
...
+        style: {
+            height: 59,
+            backgroundColor: '#2FAFCC',
+        },
+        indicatorStyle: {
+            backgroundColor: 'transparent',
+        },
+        upperCaseLabel: false,
+        labelStyle: {
+            color: '#FFFFFF',
```
santee: since you've started styling, put all style changes to stylesheet (https://facebook.github.io/react-native/docs/stylesheet.html). Previous 20/30 were here just for the sake of poc

mobile/src/layout/root-navigator.tsx
```diff
@@ -9,26 +9,59 @@ import { TabNavigationOptionsFactory } from './tab-navigation-options-factory';
...
+            path: '/',
+            navigationOptions: navOptionsFactory.create('Home', require('../../src/images/inactive-tab-bar/News.png'), require('../../src/images/inactive-tab-bar/News.png'))
+        },
+        Feed: {
+            screen: FeedScreen,
+            path: '/feeds',
+            navigationOptions: navOptionsFactory.create('People', require('../../src/images/inactive-tab-bar/People.png'), require('../../src/images/inactive-tab-bar/People.png'))
+        },
+        Organization: {
+            screen: OrganizationScreen,
```
santee: why organizationscreen has helpdesk as an icon, why do you have `Management2` and so on. Clean it up

arcelena:
Images.
I think the way when separate folder "images" with different folders of current images inside is more comfortable way to manage images in the project. Whould u prefer to put them with components?

mobile/src/layout/root-navigator.tsx
arcelena: Ok, sure i will do it.

mobile/src/layout/root-navigator.tsx
arcelena: It is cause all changes now contains just icons and styles according to mockup, navigation to current screens will be fixed later.

mobile/src/layout/tab-navigation-options-factory.tsx
arcelena: It is type for image source(like string, number and etc).

arcelena:
simulator screen shot - iphone 8 plus - 2017-12-14 at 11 01 35

PR #27, opened 2018-01-21
Feature/calendar days tiles

serjKim:
Implemented days component which looks like tiles


mobile/src/calendar/calendar-screen.tsx
```diff
@@ -1,12 +1,13 @@
...
 import styles from '../layout/styles';
+import { Days } from './days/days';
 
 export class CalendarScreen extends Component {
 
     public render() {
-        return <View style={styles.container}>
-            <Text>Calendar</Text>
+        return <View>
+            <Days />
```
santee: DaysCounters maybe?

mobile/src/calendar/calendar-screen.tsx
```diff
@@ -1,12 +1,13 @@
 import React, { Component } from 'react';
 import { View, Text } from 'react-native';
 import styles from '../layout/styles';
+import { Days } from './days/days';
 
 export class CalendarScreen extends Component {
```
santee: what do you think about renaming the whole folder to `Schedule` or `EmployeeCalendar`? I'm confused how you're going to name the calendar part of the calendar screen :) But it's probably better to leave it as is, since the screen component is named calendar-screen, not just calendar...

mobile/src/calendar/days/days.tsx
```diff
@@ -0,0 +1,58 @@
...
+class DaysImpl extends Component<DaysProps & DaysDispatchProps, {}> {
+    public componentDidMount() {
+        this.props.loadDays();
+    }
+
+    public render() {
+        const { vacation, off, sick } = this.props.days;
+
+        const vacationTile = vacation
+            ? <Tile leftDays={vacation.leftDays} allDays={vacation.allDays} title={vacation.title} leftColor={colors.days.left} allColor={colors.days.all} />
```
santee: One thing I'm not sure I like is `Tile` name. It is too generic, and if we add tiles somewhere else we'll be confused while importing things... On the other hands, they are encapsulated inside the folder and having DaysTile / DaysCounter/ DaysTileIndicator is very messy. Do not change anything unless you have a better idea :)

mobile/src/calendar/days/tile/tile-indicator.tsx
```diff
@@ -0,0 +1,45 @@
...
+        return (
+            <View style={styles.tileIndicator}>
+                <View style={this.styles.left} />
+                <View style={this.styles.all} />
+            </View>
+        );
+    }
+
+    private calculateStyles() {
+        this.styles = StyleSheet.create({
```
santee: couldn't you make it work without this magic? :)

mobile/src/calendar/days/tile/tile.tsx
```diff
@@ -0,0 +1,44 @@
...
+        return (
+            <View style={styles.tile}>
+                <View style={styles.tileContent}>
+                    <Text>
+                        <Text style={styles.tileLeftDays}>{this.props.leftDays}</Text>
+                        { this.props.showAllDays ? <Text style={styles.tileAllDays}> / {this.props.allDays}</Text> : null }
+                    </Text>
+                    <Text style={styles.tileTitle}>{this.props.title}</Text>
+                </View>
+                <TileIndicator value={indicatorValue} leftColor={this.props.leftColor} allColor={this.props.allColor} />
```
santee: this is confusing. Why tile indicator value should be percents and that does it even mean? It's not clear from api. Woudn't percentsFilled (or whatever you do with this value) be a better name?

mobile/src/calendar/days/days.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { Component } from 'react';
+import { View, Text, StyleSheet } from 'react-native';
+import { styles, colors } from '../styles';
+import { Tile, EmptyTile } from './tile/tile';
+import { TileSeparator } from './tile/tile-separator';
+import { Days as DaysModel } from '../../reducers/calendar/days.model';
+import { AppState } from '../../reducers/app.reducer';
+import { Dispatch, connect } from 'react-redux';
+import { CalendarActions, loadDays, LoadDays } from '../../reducers/calendar/calendar.action';
+import { bindActionCreators } from 'redux';
```
santee: do you use it anywhere?

mobile/src/reducers/calendar/calendar.reducer.ts
```diff
@@ -0,0 +1,30 @@
...
+import { DaysItem, Days } from './days.model';
+import { Reducer } from 'redux';
+import { CalendarActions } from './calendar.action';
+
+export const daysReducer: Reducer<Days> = (state = { vacation: null, off: null, sick: null }, action: CalendarActions) => {
+    switch (action.type) {
+        case 'LOAD-DAYS-FINISHED':
+            return {
+                vacation: action.days.vacation,
+                off: action.days.off,
```
santee: don't save bytes. `daysOff`, `vacationDaysLeft`. And what is `sick` supposed to represent? :) Days until sicklist is closed? Then call it appropriately :)

mobile/src/reducers/calendar/days.model.ts
```diff
@@ -0,0 +1,31 @@
+import { dataMember, required } from 'santee-dcts';
+
+// TODO: mock models
+
+export class DaysItem {
```
santee: well, I see the structure needs to be revised :) I'll start working on API

mobile/src/calendar/days/days.tsx
serjKim: I relied on folder structure, it would separate those generic named components, but I'm not sure too. Maybe that:
````
DaysCounters\ (agree, DaysCounters sounds good)
     DaysCounter.tsx
     DaysCounterIndicator.tsx
     DaysCounterSeparator.tsx
````

mobile/src/calendar/days/tile/tile.tsx
serjKim: ok, I'll rename to ```percentageValue```

mobile/src/calendar/days/days.tsx
serjKim: Removed

mobile/src/calendar/days/tile/tile-indicator.tsx
serjKim: Removed

mobile/src/reducers/calendar/calendar.reducer.ts
serjKim: Fixed

PR #29, opened 2018-01-23
Feature/calendar api

santee:
@serjKim just a first idea for days off. I've added counters directly to employee model, couldn't come up with a better idea, so you need to request it every time after you submit a dayoff and recieve a successful response

### DIFF
--- a/server/Arcadia.Assistant.Web/Controllers/EmployeesController.cs
+++ b/server/Arcadia.Assistant.Web/Controllers/EmployeesController.cs
@@ -75,6 +75,8 @@
                         var employee = EmployeeModel.FromMetadata(x.Metadata);
                         var photo = await x.Actor.Ask<GetPhoto.Response>(GetPhoto.Instance, timeout, token);
                         employee.Photo = photo.Photo;
+                        employee.HoursCredit = 12;
+                        employee.VacationDaysLeft = 28;
                         return employee;
                     });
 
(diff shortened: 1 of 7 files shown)


server/Arcadia.Assistant.Web/Controllers/EmployeesController.cs
```diff
@@ -75,6 +75,8 @@ public EmployeesController(IActorRefFactory actorSystem, ActorPathsBuilder paths
                         var employee = EmployeeModel.FromMetadata(x.Metadata);
                         var photo = await x.Actor.Ask<GetPhoto.Response>(GetPhoto.Instance, timeout, token);
                         employee.Photo = photo.Photo;
+                        employee.HoursCredit = 12;
```
serjKim: Did you forget to commit this property?

santee: no, the property is committed, name change didn't pass through though...

santee: fixed

server/Arcadia.Assistant.Web/Models/EmployeeModel.cs
```diff
@@ -26,6 +26,14 @@ public class EmployeeModel
 
         public DateTime? HireDate { get; set; }
 
+        public int VacationDaysLeft { get; set; }
+
+        /// <summary>
+        /// Positive values means that these days must be worked out.
+        /// Negative means that these can be taked as days off
+        /// </summary>
+        public int HoursCredit { get; set; }
```
serjKim: Maybe versa vice? positive means days off, negative - work out?

PR #31, opened 2018-01-25
Closes #17

### LINKED ISSUE #17: Create basic photo component
Use Image + base64-string from employee.photo class

victor-kotov-arcadia:
Closes #17 Create basic photo component

Basic photo component means an employee’s avatar photo with rounded and
bordered style applied according to the app’s mockups.


mobile/src/common/rounded-avatar.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { Component } from 'react';
+import { View, Image } from 'react-native';
+
+export interface Props {
+    photoBase64: string;
```
santee: where does mime type go? Do you suggest to merge it in base64 outside of the component?

mobile/src/common/rounded-avatar.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { Component } from 'react';
+import { View, Image } from 'react-native';
+
+export interface Props {
```
santee: rename Props please to something like RoundedAvatarProps. Otherwise it'll be harder to find this type or export it (what if different file exports this Props and something called Props from another file?)

mobile/src/common/index.tsx
```diff
@@ -0,0 +1 @@
+export { RoundedAvatar } from './rounded-avatar';
```
santee: We've discussed it already, having index file doesn't make development much easier but adds some problems with export, navigation and refactoring, so please do not use them.

mobile/src/common/rounded-avatar.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { Component } from 'react';
```
santee: `common` is a shortcut for `i-do-not-know-there-to-put-my-things-folder` :). As this is an avatar, it's supposed to be somewhere in `people` folder

mobile/src/common/index.tsx
victor-kotov-arcadia: Yes, we'd discussed it previously, but I don't see any issues with the component usage/import in practice, so does it really happen? Could you please give an example?

santee: Sure.
Typically if you print, say, `Round...` the auto-complete suggests you to import something from the list of found literals. However, in that case there are two *different* literals that reside in different files, so it gives you a choice of two. It also gives you a problem with `go to definition` because it sends you to the index file instead of the source.
Sometimes it happens that some objects from the folder are imported from the index file whereas others are imported from specific files which adds to confusion.
For some autocomplete tools default option is to export it directly from the original file, developers have to go and change paths manually and people tend to forget to do that if there are no automatic tools that check that. There are many other small inconveniences, i won't describe all of them.
On top of that, using 'index' files can create some problems during compilation (it depends on `module` setting in tsconfig file. At the moment, react-native asks for the module type that doesn't suffer from that, but it might change in future version (although i doubt that :) ).

We've tried this approach on some older projects, faced that problems and decided to give it away, so we do not use it anymore.

mobile/src/common/rounded-avatar.tsx
victor-kotov-arcadia: I see it as a place for components what could be used elsewhere not at some specific screen only. For example, a button with a specific style what could appear in forms/screen across the app would be also placed here.

mobile/src/common/rounded-avatar.tsx
```diff
@@ -0,0 +1,26 @@
...
+    innerBorderWidth: number;
+}
+
+export class RoundedAvatar extends Component<RoundedAvatarProps> {
+    public render() {
+        const { outerBorderColor, innerBorderColor, photoBase64, avatarSize, outerBorderWidth, innerBorderWidth } = this.props;
+        const imageSize = avatarSize - outerBorderWidth;
+
+        return (
+            <View style={{ height: avatarSize, width: avatarSize, justifyContent: 'center', alignItems: 'center', borderRadius: avatarSize * 0.5, borderColor: outerBorderColor, borderWidth: outerBorderWidth }}>
```
serjKim: Could you move your styles to the separate file `styles.ts`?

mobile/src/common/index.tsx
serjKim: if we want to have Web version in the future, we will try to reuse some code parts, and we will probably use Webpack on client side. Webpack handles bad index.ts (and treeshaking probably won't work properly) :smile:

mobile/src/common/rounded-avatar.tsx
```diff
@@ -12,14 +12,14 @@ export interface Props {
...
-export class RoundedAvatar extends Component<Props> {
+export class RoundedAvatar extends Component<RoundedAvatarProps> {
     public render() {
         const { outerBorderColor, innerBorderColor, photoBase64, avatarSize, outerBorderWidth, innerBorderWidth } = this.props;
         const imageSize = avatarSize - outerBorderWidth;
 
         return (
             <View style={{ height: avatarSize, width: avatarSize, justifyContent: 'center', alignItems: 'center', borderRadius: avatarSize * 0.5, borderColor: outerBorderColor, borderWidth: outerBorderWidth }}>
-                <Image source={{uri: photoBase64}} style={{ height: imageSize, width: imageSize, borderRadius: imageSize * 0.5, borderColor: innerBorderColor, borderWidth: innerBorderWidth }} />
+                <Image source={{uri: 'data:image/png;base64,' + photoBase64}} style={{ height: imageSize, width: imageSize, borderRadius: imageSize * 0.5, borderColor: innerBorderColor, borderWidth: innerBorderWidth }} />
```
santee: I thought that mime type should be provided by the parent component, not hardcoded. Server API returns that as a separate field in Photo model

mobile/src/common/index.tsx
victor-kotov-arcadia: ok, we already agreed do not use it, thanks for details

mobile/src/common/rounded-avatar.tsx
victor-kotov-arcadia: ok, missed this attribute, it will be used

mobile/src/common/rounded-avatar.tsx
serjKim: or just put it next to your component

PR #32, opened 2018-01-25
Feature/load user

serjKim:


mobile/src/reducers/organization/organization.epics.ts
```diff
@@ -1,13 +1,36 @@
...
+// TODO: Handle error, display some big alert blocking app...
+export const loadEmployeeForUserEpic$ = (action$: ActionsObservable<LoadEmployeeForUser>) =>
+    action$.ofType('LOAD-EMPLOYEE-FOR-USER')
+        .switchMap(x => ajaxGetJSON(`${url}/employees/${x.user.employeeId}`))
+        .map(x => deserialize(x, Employee))
+        .map(x => loadEmployeeForUserFinished(x));
+
+export const loadEmployeeForUserFinishedEpic$ = (action$: ActionsObservable<LoadEmployeeForUserFinished>) =>
+    action$.ofType('LOAD-EMPLOYEE-FOR-USER-FINISHED')
+        .map(x => loadDepartments());
```
serjKim: The point was to prevent any requests until the user has been loaded. What do you think? Looks very implicit.

santee: why would you want that? User endpoint response is not related to authentication anyhow, and this will be handled separately.

mobile/src/reducers/organization/organization.epics.ts
```diff
@@ -1,15 +1,41 @@
...
+        .map(x => deserialize(x, User))
+        .map(x => loadUserFinished(x));
+
+export const loadUserFinishedEpic$ = (action$: ActionsObservable<LoadUserFinished>) =>
+    action$.ofType('LOAD-USER-FINISHED')
+        .map(x => loadEmployeeForUser(x.user));
+
+// TODO: Handle error, display some big alert blocking app...
+export const loadEmployeeForUserEpic$ = (action$: ActionsObservable<LoadEmployeeForUser>) =>
+    action$.ofType('LOAD-EMPLOYEE-FOR-USER')
```
santee: don't we already have an action that loads data from this endpoint?

serjKim: we have only the action, but don't have epic.

mobile/src/reducers/organization/organization.reducer.ts
```diff
@@ -4,20 +4,32 @@ import { combineReducers } from 'redux';
...
+    loadEmployeesForDepartmentEpic$ as any,
+    loadUserEpic$ as any,
+    loadUserFinishedEpic$ as any,
+    loadEmployeeEpic$ as any);
 
 export const organizationReducer = combineReducers<OrganizationState>({
     departments: departmentsReducer,
-    employees: employeesReducer
+    employees: employeesReducer,
+    user: userReducer
```
santee: I do not think that user belongs under organization. Should it be in a folder of its own? We'll have to support some settings later, for example.

mobile/src/reducers/organization/user.reducer.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const initState: UserState = {
+    user: null,
+    employee: null
+};
+
+export const userReducer: Reducer<UserState> = (state = initState, action: OrganizationActions) => {
+    switch (action.type) {
+        case 'LOAD-USER-FINISHED':
+            return {
```
santee: hint: you can use `return { ...state, user: action.user} instead. That saves you from forgetting to copy new properties when they're introduced

mobile/src/reducers/organization/organization.epics.ts
```diff
@@ -1,15 +1,39 @@
...
+        .map(x => loadEmployee(x.user.employeeId));
+
+export const loadEmployeeEpic$ = (action$: ActionsObservable<LoadEmployee>) =>
+    action$.ofType('LOAD_EMPLOYEE')
+        .groupBy(x => x.employeeId)
+        .map(x =>
+            x.switchMap(y => ajaxGetJSON(`${url}/employees/${y.employeeId}`)).map(obj => deserialize(obj, Employee))
+        )
+        .mergeAll()
+        .map(x => loadEmployeeFinished(x));
```
santee: you forgot a catch.

mobile/src/reducers/organization/organization.epics.ts
```diff
@@ -26,7 +50,6 @@ export const loadChiefsEpic$ = (action$: ActionsObservable<LoadDepartmentsFinish
         .map(x => loadEmployeeFinished(x))
```
santee: can we probably edit loadChiefsEpic so it maps to series of 'loadEmployee' actions?

mobile/src/reducers/organization/organization.epics.ts
serjKim: Fixed

mobile/src/reducers/organization/organization.epics.ts
serjKim: Fixed

mobile/src/reducers/organization/organization.reducer.ts
serjKim: Fixed

mobile/src/reducers/organization/user.reducer.ts
serjKim: Right. Fixed

PR #35, opened 2018-01-30
Feature/basic photocomponent

victor-kotov-arcadia:
Component code updated to use calculated values for photo and container.


mobile/src/people/avatar.tsx
```diff
@@ -5,45 +5,44 @@ const styles = StyleSheet.create({
...
-    // or some other color - refer to design.
-    outerFrameStyle: ViewStyle;
-    imageStyle: ImageStyle;
+    style: ViewStyle;
 }
 
 export class Avatar extends Component<AvatarProps> {
+    private borderRadius: number;
+    
+    public onLayout = (e: any) => {
```
serjKim: ```e: LayoutChangeEvent```

mobile/src/people/avatar.tsx
```diff
@@ -1,49 +1,48 @@
...
-    outerFrameStyle: ViewStyle;
-    imageStyle: ImageStyle;
+    style: ViewStyle;
 }
 
 export class Avatar extends Component<AvatarProps> {
+    private borderRadius: number;
+    
+    public onLayout = (e: LayoutChangeEvent) => {
+        this.borderRadius = e.nativeEvent.layout.width * 0.5;
```
santee: any reason why `this.borderRadius` is a field instead of a parameter in 'setState'?

santee: it actually depends on call order whether you'll get a valid radius in render() method or not... https://reactjs.org/docs/faq-state.html

PR #40, opened 2018-02-08
Century gothic font

serjKim:
Add Century gothic font.  
Fix font sizes according to Figma's sketches.
Verified on andriod emulator and iphone x.


demkalkov:
Since we can output text only with `Text` element, wouldn't it be better to create our own text component that will extend standard `Text` and will have the right font family?

serjKim:
> Since we can output text only with Text element, wouldn't it be better to create our own text component that will extend standard Text and will have the right font family?

serjKim:
> Since we can output text only with Text element, wouldn't it be better to create our own text component that will extend standard Text and will have the right font family?

Maybe right. But sometimes it's impossible to use our component (for example NavBar title), and we would have to write fontFamily manually.

Would it better just spread object styles?:
```
const globalFontStyles = { fontFamily: ...  }

const someStyles = {
...globalFontStyles,
...
};
```

demkalkov:
We may fix nav text issue in the future, and assign styles manually for now. But in most cases you have to use `Text` and it is a bad idea to write font-family or import and  add global font styles for every singe text element

serjKim:
Fixed

mobile/package.json
```diff
@@ -59,5 +59,10 @@
 			"!artifacts/**/*.spec.js",
 			"!artifacts/**/*.index.js"
 		]
+	},
+	"rnpm": {
```
santee: dafaq is this?

mobile/src/override/acradia-text.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { Component } from 'react';
+import { TextProperties, StyleSheet, Text, ViewStyle, TextStyle } from 'react-native';
+
+const commonStyles = StyleSheet.create({
+    text: {
+        fontFamily: 'CenturyGothic'
+    }
+});
+
+export class ArcadiaText extends Component<TextProperties> {
```
santee: really????? is this a way to go???

santee: Yep, read the documentation. Spilled my smoothies.

santee: Check out `Text.defaultProps.style = { fontFamily: 'some-font' }` ?

mobile/package.json
serjKim: react-native cli copies the assets to the projects from this path

mobile/src/override/acradia-text.tsx
santee: You all guys support using ArcadiaText component instead of defaultProps?

serjKim: It will be rewritten, if I pass some styles

serjKim: My code merges objects

serjKim: > You all guys support using ArcadiaText component instead of defaultProps?

I would prefer to merge styles than use this component )

mobile/package.json
serjKim: I've added it manually that rnpm copies it to the project

santee: this is what I'm asking. Why? all other fonts work without rnpm

serjKim: This is a custom font. To use it should be added to the project as resource. rnpm does it at once to both platforms. is there another way?

mobile/src/override/acradia-text.tsx
serjKim: So, we've decided to call StyledText

PR #45, opened 2018-02-14
tab navigation vertical align and transparent icons

demkalkov:


santee:
This is not needed, see @arcelena 's PR

PR #47, opened 2018-02-14
Use ScrollView on a Profile component to allow scrolling when the con…

demkalkov:
…tent is too high for the screen

### DIFF
--- a/mobile/src/profile/profile.tsx
+++ b/mobile/src/profile/profile.tsx
@@ -1,8 +1,7 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
-import { View, LayoutChangeEvent, Text, Image, ImageStyle, StyleSheet } from 'react-native';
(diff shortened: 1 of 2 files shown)


mobile/src/profile/profile.tsx
```diff
@@ -30,7 +29,8 @@ export class Profile extends Component<ProfileProps> {
         const tiles = this.getTiles(employee);
         const contacts = this.getContacts(employee);
 
-        return <View style={styles.container}>
+        return (
+        <ScrollView style={layoutStyles.scrollView}>
```
santee: formatting got screwed. Any why parenthesis?

demkalkov: Because you can't just move element to the new line without wrapping it

demkalkov: @santee ?

PR #49, opened 2018-02-14
noInnerBorder prop added

victor-kotov-arcadia:
- noInnerBorder prop added
- image border calculation refined according to the design


demkalkov:
Now I don't see any issues too. @santee

PR #53, opened 2018-02-16
copy images to artifacts

demkalkov:


mobile/src/people/avatar.tsx
```diff
@@ -53,7 +53,7 @@ export class Avatar extends Component<AvatarProps, AvatarState> {
         const mimeType = this.validateMimeType(this.props.photo);
         const photoBase64 = this.validateEncodedImage(this.props.photo);
 
-        const image = !mimeType || !photoBase64 ? require('../../src/people/userpic.png') : { uri: mimeType + photoBase64 };
+        const image = !mimeType || !photoBase64 ? require('./userpic.png') : { uri: mimeType + photoBase64 };
 
         const outerFrameFlattenStyle = StyleSheet.flatten([
```
arcelena: I think all styles in this file should be replaced to separate  file.

mobile/yarn.lock
```diff
@@ -3148,6 +3130,10 @@ natural-compare@^1.4.0:
   version "1.4.0"
   resolved "https://registry.yarnpkg.com/natural-compare/-/natural-compare-1.4.0.tgz#4abebfeed7541f2c27acfb29bdbbd15c8d5ba4f7"
 
+ncp@^2.0.0:
+  version "2.0.0"
+  resolved "http://cvddev.cloudapp.net/artifactory/api/npm/covance-npm/ncp/-/ncp-2.0.0.tgz#195a21d6c46e361d2fb1281ba38b91e9df7bdbb3"
```
santee: lol

mobile/src/people/avatar.tsx
arcelena: Lines 5-21 is not calculated from props and can be replaced. If PR is not about it and have no permission to replace it - ok.

PR #55, opened 2018-02-16
profile contacts linking

demkalkov:


mobile/src/profile/profile.tsx
```diff
@@ -118,26 +118,34 @@ export class Profile extends Component<ProfileProps> {
...
+                        <StyledText style={contactStyles.text}>{contact.text}</StyledText>
+                    </View>
                 </View>
-            </View>
+            </TouchableOpacity>
         ));
     }
+
+    private openLink(url: string) {
+        return () => Linking.openURL(url).catch(err => null); //Should we show any exception here?
```
santee: console.error(err) should be enough to start with

PR #58, opened 2018-02-19
Feature/clean up profile design

arcelena:


mobile/src/profile/styles.ts
```diff
@@ -87,30 +90,31 @@ export const contentStyles = StyleSheet.create({
...
-        justifyContent: 'center'
     }
 });
 
 export const tileStyles = StyleSheet.create({
     container: {
-        padding: 1,
-        width: '25%'
+        padding: paddingTile,
+        width: (Dimensions.get('window').width - 2 * paddingInfoContainer - 3 * paddingTile) / 4,
```
serjKim: it's really complicated.

What if instead of ```padding: paddingTile``` you try to add separate view with ```width: paddingTile``` between your tiles:

```javascript
        const TitleSeparator = () => <View style={{flexBasis: paddingTile /*initial size (width)*/, backgroundColor: '#fff'}}></View>;
...

        const lastIndex = tilesData.length - 1;

        return tilesData.map((tile, index) => (
            <React.Fragment>
                <View key={tile.label} style={tileStyles.container}>
                    <View style={tileStyles.tile}>
                        <View style={tileStyles.iconContainer}>
                            <ApplicationIcon name={tile.icon} size={tile.size} style={tile.style} />
                        </View>
                        <StyledText style={tileStyles.text}>{tile.label}</StyledText>
                    </View>
                </View>
                {
                    lastIndex !== index 
                        ? <TitleSeparator key={`${tile.label}-${index}`} />
                        : null
                }
            </React.Fragment>
        ));
```

mobile/src/profile/styles.ts
```diff
@@ -138,10 +142,13 @@ export const tileStyles = StyleSheet.create({
 export const contactStyles = StyleSheet.create({
     container: {
         flexDirection: 'row',
-        paddingBottom: 10
+       paddingLeft: 45,
```
demkalkov: Are you trying to align contacts with padding?

mobile/src/profile/styles.ts
demkalkov: What is wrong with 25%?

serjKim:
Did you check on Android?

mobile/src/profile/styles.ts
arcelena: I align left contacts, why not?

mobile/src/profile/styles.ts
arcelena: The problem is with pixels between tiles, i will try what Sergey has suggested.

arcelena: i will try

mobile/src/profile/styles.ts
demkalkov: Because it looks like centered
capture

serjKim:
I see on my android emulator:

arcelena:
It seems scrollview does not work on android sim.

mobile/src/profile/styles.ts
arcelena: Our designer (Ilya) said it should be left align.

demkalkov:
Well it used to work.
Anyway @santee has told to remove it completely as far as I remember

santee:
can it only scroll down? so we do not have this bikini effect? or just put a blue background...

mobile/src/profile/styles.ts
```diff
@@ -57,16 +60,16 @@ export const layoutStyles = StyleSheet.create({
     content: {
         flex: 1,
         flexDirection: 'column',
-        marginTop: placeholderHeight
-
+        marginTop: circleDiameter * 0.5 - headerGap,
+        height: 350
```
serjKim: can we use flex instead of hardcode?

arcelena: No, cause flex  does not work good for scrollView on Android.

mobile/src/profile/styles.ts
```diff
@@ -87,29 +90,32 @@ export const contentStyles = StyleSheet.create({
     position: {
         fontSize: 13,
         textAlign: 'center',
-        color: '#000'
+        color: '#000',
+        top: 10
```
serjKim: I just wonder how does `top` work without position relative/absolute

PR #59, opened 2018-02-20
calendar redesign

demkalkov:


mobile/src/calendar/calendar-legend.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+    private getVisibleCalendarEvents() {
+        const { intervals, weeks } = this.props;
+        const calendarEventsMap: Map<CalendarEventsType, number> = new Map();
+
+        if (!weeks || !intervals) {
+            return calendarEventsMap;
+        }
+
+        weeks.forEach(week => {
```
serjKim: I would put this logic to ```IntervalsModel```, for example to ```eventTypesByMonth()```

mobile/src/tabbar/tab-navigator.tsx
```diff
@@ -13,6 +13,7 @@ const tabNavigatorConfig: TabNavigatorConfig = {
         showIcon: true,
     },
     animationEnabled: false,
+    initialRouteName: 'Calendar'
```
serjKim: Revert )

mobile/src/calendar/styles.ts
```diff
@@ -119,74 +140,65 @@ export const agendaStyles = StyleSheet.create({
...
-export const daysCounterStyles = StyleSheet.create({
+export const selectedDayStyles = StyleSheet.create({
     container: {
-        backgroundColor: daysCounterPrimaryColor,
-        flexDirection: 'column',
-        flex: 1
-    },
-    content: {
-        flexDirection: 'column',
+        //transform: [{ translateX: -(circleDiameter / 2) }],
```
serjKim: Remove commented code

serjKim:
Merge to master

mobile/src/calendar/styles.ts
```diff
@@ -83,18 +83,39 @@ export const calendarStyles = StyleSheet.create({
...
+
+            default:
+                return null;
+        }
+    }
+}
 
 // serjKim: Depends on count of overlapped intervals?..
-const intervalOpacity = 0.5;
+const intervalOpacity = .95;
```
serjKim: ?

demkalkov: When opacity is .5 the colors are too pale and do not match the design at all

mobile/src/calendar/calendar-legend.tsx
demkalkov: Isn't it view specific logic that might not be required anywhere else?

mobile/src/calendar/styles.ts
serjKim: But it's difficult to determine when some overlapped interval is started or finished

mobile/src/calendar/calendar-legend.tsx
serjKim: I think it looks like a part of IntervalsModel api. And doesn't depend on whether you are going to reuse it or not, having it in component is difficult to test it.

demkalkov: @santee

mobile/src/calendar/calendar-legend.tsx
```diff
@@ -0,0 +1,66 @@
...
+                <View>
+                    {legend}
+                </View>
+            </View>
+        );
+    }
+
+    private getVisibleCalendarEvents() {
+        const { intervals, weeks } = this.props;
+        const calendarEventsMap: Map<CalendarEventsType, number> = new Map();
```
santee: can you put generic parameters in constructor call?

mobile/src/calendar/today.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { View } from 'react-native';
+import { connect } from 'react-redux';
+import { DayModel } from '../reducers/calendar/calendar.model';
+import { SelectedDay } from './days-counters/selected-day';
+import { AppState } from '../reducers/app.reducer';
+
+interface TodayProps {
+    selectedCalendarDay: DayModel;
+}
+class TodayImpl extends Component<TodayProps> {
```
santee: Should be removed

PR #60, opened 2018-02-20
Feature/toptabbar people employees

victor-kotov-arcadia:
People screen populated with data at Department and Room tabs.
Employees list ordered by employee's name.


mobile/src/people/employees-list-item.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { StyledText } from '../override/styled-text';
+
+interface EmployeesListItemProps {
+    employee: Employee;
+    id: string;
+}
+
+export class EmployeesListItem extends React.Component<EmployeesListItemProps> {
+    public render() {
+        const employeeName = this.props.employee ? this.props.employee.name : 'Unknown';
```
demkalkov: Not `Unknown`. null

demkalkov: Are there any employee in a db without a name?

mobile/src/people/home-people-screen.tsx
```diff
@@ -18,10 +19,13 @@ const navBar =  new TopNavBar('People');
...
 class HomePeopleScreenImpl extends React.Component<PeopleScreenProps> {
     public static navigationOptions = navBar.configurate();
-
+    
     public render() {
+        
         return (
-            <Text>PeopleScreen. There are { this.props.departments.length} departments</Text>
+            <View style={{ flex: 1 }}>
+                <TopTabBarNavigator />
```
demkalkov: Do you really need a component that only returns another one?

mobile/src/people/employees-list-item.tsx
santee: should not be any. I think such employees fill throw an exception on deserialization

mobile/src/people/employees-list.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { StyledText } from '../override/styled-text';
+
+export interface EmployeesListProps {
+    employees: Employee[];
+}
+
+export class EmployeesList extends React.Component<EmployeesListProps> {
+    public render() {
+        const employees = this.props.employees.sort((a, b) => {
+            if (a.name < b.name) {
```
santee: ? why does it return 0 if a.name > b.name?

mobile/src/people/home-people-screen.tsx
santee: are you referring to the whole file? I think yes, I'd rather have it as a wrapper. However, it should not have Props at all, they are not used anywhere.

santee: TopTabBarNavigator should not be called as such. It works only in scope of peoples scene. Rename it to PeopleScreenNavigator or something like that.

mobile/src/people/people-department.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { EmployeesList } from './employees-list';
+import { AppState } from '../reducers/app.reducer';
+import { PeopleActions, navigatePeopleDepartment } from '../reducers/people/people.action';
+import { loadEmployeesForDepartment } from '../reducers/organization/organization.action';
+import { EmployeesStore, EmployeeMap, EmployeeIdsGroupMap } from '../reducers/organization/employees.reducer';
+
+interface PeopleDepartmentProps {
+    employeesMap: EmployeeMap;
+    departmentId: string;
+    employeesSubsetFilterCallback: any;
```
santee: type it better please, no anies

mobile/src/people/people-department.tsx
```diff
@@ -0,0 +1,45 @@
...
+const mapStateToProps = (state: AppState): PeopleDepartmentProps => ({
+    employeesMap: state.organization.employees.employeesById,
+    departmentId: state.userInfo.employee.departmentId,
+    employeesSubsetFilterCallback: state.people.employeesDepartmentSubsetFilterCallback
+});
+
+const mapDispatchToProps = (dispatch: Dispatch<PeopleActions>) => ({
+    navigatePeopleDepartment: (departmentId: string) => { 
+        dispatch(navigatePeopleDepartment(departmentId)); 
+        dispatch(loadEmployeesForDepartment(departmentId));
```
santee: the second dispatch should be called in response to the first one, something like action.ofType('NAVIGATE_DEPARTMENT).map(x => loadEmployeesForDepartment(x.departmentId))

mobile/src/people/employees-list.tsx
santee: i guess you forgot .name in second clause. Also, please put it a separate file / function called employeeComparer(first, second).

mobile/src/people/people-room.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const mapStateToProps = (state: AppState): PeopleRoomProps => ({
+    employeesMap: state.organization.employees.employeesById,
+    roomNumber: state.userInfo.employee.roomNumber,
+    employeesSubsetFilterCallback: state.people.employeesRoomSubsetFilterCallback
+});
+
+const mapDispatchToProps = (dispatch: Dispatch<PeopleActions>) => ({
+    navigatePeopleRoom: (roomNumber: string) => { 
+        dispatch(navigatePeopleRoom(roomNumber));
```
santee: same thing with actions mapping

mobile/src/reducers/people/people.reducer.ts
```diff
@@ -0,0 +1,48 @@
...
+export interface PeopleState {
+    employees: EmployeeMap;
+    employeesDepartmentSubsetFilterCallback: any;
+    employeesRoomSubsetFilterCallback: any;
+    employeesSubsetFilterCallback: any;
+}
+
+const initState: PeopleState = {
+    employees: Map(),
+    employeesDepartmentSubsetFilterCallback: function(employee: Employee) {
```
santee: just a filter please, without Callback. It's just not in the js tradition of naming :) You may call it Predicate instead of FilterCallback, that's more acceptable in CS-world :)

Also, here is your typing (employee: Employee) => boolean;

mobile/src/reducers/people/people.reducer.ts
```diff
@@ -0,0 +1,48 @@
...
+import { User } from '../user/user.model';
+import { Employee } from '../organization/employee.model';
+import { OrganizationActions, loadEmployeesForDepartment } from '../organization/organization.action';
+import { UserActions } from '../user/user.action';
+import { PeopleActions } from './people.action';
+import { EmployeeMap } from '../organization/employees.reducer';
+
+export interface PeopleState {
+    employees: EmployeeMap;
+    employeesDepartmentSubsetFilterCallback: any;
```
santee: hold on. Why there are three different functions? There should be only one and you switch between them. Then you can reuse the same Component for Room and Department screens, the only thing that will change is a function returned.

mobile/src/reducers/people/people.reducer.ts
```diff
@@ -0,0 +1,48 @@
...
+    },
+    employeesSubsetFilterCallback: function(employee: Employee) {
+        return employee.employeeId === '';
+    }
+};
+
+export const peopleReducer: Reducer<PeopleState> = (state = initState, action: PeopleActions) => {
+    switch (action.type) {
+        case 'NAVIGATE-PEOPLE-DEPARTMENT':
+            return {...state, employeesDepartmentSubsetFilterCallback: function(employee: Employee) {
```
santee: Instead of using anonymous function here, please move this functions above in the file. The reason is that each time you get here it returns a new instance of the function, which will cause a rerender in components (if, you, say, call this event multiple times being on the same screen... which can occur)

mobile/src/reducers/user/user-info.reducer.ts
```diff
@@ -1,3 +1,4 @@
+import { Map, Set } from 'immutable';
```
santee: Please revert the whole file

mobile/src/reducers/user/user.action.ts
```diff
@@ -1,5 +1,6 @@
 import { User } from './user.model';
 import { Employee } from '../organization/employee.model';
+import { EmployeeMap } from '../organization/employees.reducer';
```
santee: revert the whole file

mobile/src/reducers/user/user.epics.ts
```diff
@@ -1,9 +1,9 @@
 import { ajaxGetJSON } from 'rxjs/observable/dom/AjaxObservable';
 import { deserialize } from 'santee-dcts/src/deserializer';
-import { loadEmployee, LoadEmployeeFinished } from '../organization/organization.action';
+import { loadEmployee, LoadEmployeeFinished, OrganizationActions, LoadEmployeesForDepartment } from '../organization/organization.action';
```
santee: Revert the whole file.

mobile/src/reducers/user/user.reducer.ts
```diff
@@ -5,4 +5,5 @@ import { loadUserEpic$, loadUserFinishedEpic$ } from './user.epics';
 
 export const userEpics = combineEpics(
     loadUserEpic$ as any,
-    loadUserFinishedEpic$ as any);
\ No newline at end of file
+    loadUserFinishedEpic$ as any);
```
santee: revert

mobile/src/toptabbar/people-top-tab-bar-model.tsx
```diff
@@ -0,0 +1,30 @@
...
+import React from 'react';
+import { PeopleDepartment } from '../people/people-department';
+import { PeopleRoom } from '../people/people-room';
+import { PeopleCompany } from '../people/people-company';
+import { EmployeesList } from '../people/employees-list';
+import { TabNavigationOptionsFactory } from '../tabbar/tab-navigation-options-factory';
+
+const navOptionsFactory = new TabNavigationOptionsFactory();
+
+const peopleTopTabBarModel: NavigationRouteConfigMap = {
```
santee: Get rid of this folder, move it under people. It's only used in people scene.

santee: Or, rather, you can leave this and styles file in this folder, but get rid of people-top-tab-bar

mobile/src/toptabbar/top-tab-bar-navigator.tsx
```diff
@@ -0,0 +1,24 @@
...
+    tabBarOptions: {
+        style: topTabBarStyles.tabBar,
+        indicatorStyle: topTabBarStyles.tabBarIndicator,
+        upperCaseLabel: false,
+        labelStyle: topTabBarStyles.tabBarLabel,
+        showIcon: false,
+    },
+};
+
+export const  TopTabBarNavigator = TabNavigator(
```
santee: this const should be moved under people folder with a different name

mobile/src/toptabbar/top-tab-bar-styles.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { StyleSheet, Platform } from 'react-native';
+
+const topTabBarStyles = StyleSheet.create({
+    tabBar: {
```
demkalkov: > tabBar: {
        height: 50,
        backgroundColor: '#2FAFCC',
        justifyContent: 'center'
    },
    tabBarIndicator: {
        backgroundColor: 'rgba(0, 0, 0, 0.4)'
    }

mobile/src/people/employees-list.tsx
victor-kotov-arcadia: ok

mobile/src/people/home-people-screen.tsx
victor-kotov-arcadia: Text - forgot to delete it

mobile/src/reducers/people/people.reducer.ts
victor-kotov-arcadia: re: Callback - it was advice from Serj ;)

mobile/src/reducers/people/people.reducer.ts
victor-kotov-arcadia: If I use one universal function - it's working in such way: once I open Room tab the content at Department tab became same - so new predicate (room specific) is applied in both cases.

santee: what

mobile/src/reducers/people/people.reducer.ts
serjKim: not literally, as an example

mobile/src/people/employees-list-item.tsx
santee: @victor-kotov-arcadia this comment is still not addressed

mobile/src/reducers/organization/organization.reducer.ts
```diff
@@ -16,7 +16,6 @@ export const organizationEpics = combineEpics(
     loadEmployeeEpic$ as any,
     loadDepartmentsEpic$ as any,
     loadChiefsEpic$ as any,
-    loadDepartmentsFinishedEpic$ as any,
     loadEmployeesForDepartmentEpic$ as any
```
santee: this is duplicated in people reducer... why does people reducer exist, let's put it in one folder/file for now.

mobile/src/reducers/user/user-info.reducer.ts
```diff
@@ -9,13 +10,14 @@ export interface UserInfoState {
 }
 
 const initState: UserInfoState = {
-    employee: null
+    employee: null,
```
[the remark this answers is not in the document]
santee: @victor-kotov-arcadia still relevant

mobile/src/reducers/organization/organization.reducer.ts
victor-kotov-arcadia: ok, is it ok to have it's code placed at employees.reducer.ts ?

santee: but you don't have a separate reducer, do you? You just some new epics. Reducers are side-effect-free functions, epics are side-effects.
Move your new epics either in organization.epics or in new file called employees.epics. In that case, however, you'll need to move everything that load employees in that file.

mobile/src/reducers/organization/organization.epics.ts
```diff
@@ -48,4 +42,23 @@ export const loadEmployeesForDepartmentEpic$ = (action$: ActionsObservable<LoadE
...
+            .flatMap(x => x.map(loadEmployeeFinished))
+            .catch((e: Error) => Observable.of(loadFailedError(e.message)));
+
+export const loadEmployeesForUserDepartmentEpic$ = (action$: ActionsObservable<LoadUserEmployeeFinished>) =>
+            action$.ofType('LOAD-USER-EMPLOYEE-FINISHED')
+                .map(x => loadEmployeesForDepartment(x.employee.departmentId));
+        
+        
+export const loadEmployeesForUserRoomEpic$ = (action$: ActionsObservable<LoadUserEmployeeFinished>) =>
+                action$.ofType('LOAD-USER-EMPLOYEE-FINISHED')
```
santee: indent.... 4 and 4

PR #66, opened 2018-02-28
Feature/calendar interval boundary

serjKim:
Calendar improvements. See the git comments.

### DIFF
--- a/mobile/src/reducers/calendar/calendar-intervals-builder.spec.ts
+++ b/mobile/src/reducers/calendar/calendar-intervals-builder.spec.ts
@@ -1,5 +1,5 @@
 import moment from 'moment';
-import { WeekModel, IntervalsModel, IntervalModel } from './calendar.model';
+import { WeekModel, IntervalsModel, IntervalModel, IntervalType } from './calendar.model';
 import { CalendarEvents, CalendarEventsType, DatesInterval } from './calendar-events.model';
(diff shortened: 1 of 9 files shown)


mobile/src/reducers/calendar/calendar-intervals-builder.spec.ts
```diff
@@ -177,27 +204,153 @@ describe('CalendarIntervalsBuilder', () => {
...
         expect(intervals[0].eventType).toBe(CalendarEventsType.SickLeave);
+        expect(intervals[0].startDate).toBe(event2.dates.startDate);
+        expect(intervals[0].endDate).toBe(event2.dates.endDate);
+    });
+
+    describe('dayoff', () => {
+
+        const testDayoff = (testedEventType: CalendarEventsType.Dayoff | CalendarEventsType.AdditionalWork) => {
+            describe(testedEventType, () => {
+                it('should return interval left boundary, if startWorkingHour between (0, 4] and finishWorkingHour is 0', () => {
```
santee: finishWorkingHour should not be less then startWorkingHour. Does it actually happen?

serjKim: As I realized, if finishWorkingHour is between 0 and 4, with finishWorkingHour = 0, it means a first half?

PR #74, opened 2018-03-12
Save new sickleave event to the server

serjKim:


mobile/src/calendar/calendar-actions-button-group.tsx
```diff
@@ -77,7 +77,7 @@ const mapStateToProps = (state: AppState): ActionButtonGroupProps => ({
 
 const mapDispatchToProps = (dispatch: Dispatch<CalendarActions>) => ({
     sickLeaveActions: {
-        claim: (startDate: Moment) => { dispatch(claimSickLeave(startDate)); },
+        claim: (startDate: Moment) => { dispatch(claimSickLeave()); },
```
santee: startDate is not needed as a parameter

mobile/src/calendar/styles.ts
```diff
@@ -111,13 +111,13 @@ export class CalendarEventsColor {
...
-            case CalendarEventsType.SickLeave:
+            case CalendarEventsType.Sickleave:
                 return CalendarEventsColor.sickLeave;
 
             case CalendarEventsType.Dayoff:
                 return CalendarEventsColor.dayoff;
 
-            case CalendarEventsType.AdditionalWork:
+            case CalendarEventsType.Workout:
                 return CalendarEventsColor.additionalWork;
```
santee: want to rename it as well?

mobile/src/reducers/calendar/calendar-events.model.ts
```diff
@@ -3,9 +3,9 @@ import { Moment } from 'moment';
 
 export enum CalendarEventsType {
     Vacation = 'Vacation',
-    SickLeave = 'Sick Leave',
+    Sickleave = 'Sickleave',
     Dayoff = 'Dayoff',
-    AdditionalWork = 'Additional Work'
+    Workout = 'Workoff'
```
santee: workout :)

mobile/src/reducers/calendar/calendar-events.reducer.ts
```diff
@@ -1,34 +1,44 @@
...
+export interface EditingOfIntervalsSubState {
     editingOfIntervals: EditingOfIntervalsState;
+}
+
+export interface DisableDaysCalendarDaysBeforeSubState {
     disableCalendarDaysBefore: DayModel;
+}
+
+export interface CalendarEventsState extends 
+IntervalsSubState,
```
santee: ident

mobile/src/reducers/calendar/calendar-events.reducer.ts
```diff
@@ -84,11 +89,13 @@ export const calendarEventsReducer: Reducer<CalendarEventsState> = (state = init
                 disableCalendarActionsButtonGroup: false
             };
         case 'SELECT-CALENDAR-DAY':
-            const newSickLeaveEventState = selectSickLeaveEndDateReducer(state, action);
+            const startSickLeaveDialog = selectStartDateSickLeaveReducer(state, action);
```
santee: are we going to have to change this method every time we add another 'select something' dialog?

mobile/src/reducers/calendar/calendar-events.reducer.ts
```diff
@@ -106,6 +113,20 @@ export const calendarEventsReducer: Reducer<CalendarEventsState> = (state = init
...
             };
+        case 'CONFIRM-START-DATE-SICK-LEAVE':
+            const confirmStartDateSickLeaveDialog = confirmStartDateSickLeaveReducer(state, action);
+
+            return {
+                ...state,
+                ...confirmStartDateSickLeaveDialog
+            };
+        case 'BACK-TO-CLAIM-SICK-LEAVE':
+            const backToClaimSickLeaveDialog = backToClaimSickLeaveReducre(state, action);
```
santee: typo

mobile/src/reducers/calendar/calendar.epics.ts
```diff
@@ -1,19 +1,30 @@
...
         .catch((e: Error) => Observable.of(loadFailedError(e.message)));
 
+export const calendarEventsSavedEpic$ = (action$: ActionsObservable<ConfirmClaimSickLeave>) =>
+    action$.ofType('CONFIRM-CLAIM-SICK-LEAVE')
+        .switchMap(x => {
+
+            return ajaxPost(
+                `${apiUrl}/employees/${x.employee.employeeId}/events`,
+                x.calendarEvents,
+                { 'Content-Type': 'application/json' }).map(() => x.employee);
```
santee: this method returns you a newly created event. Why do you not use it? I think you'll need its id, for example.

santee: also, the whole switch method probably needs to be moved to another funciton, or used in a pipe

mobile/src/reducers/calendar/event-dialog/event-dialog.model.ts
```diff
@@ -56,6 +64,39 @@ export class EventDialogEmptyModel extends EventDialogModel<EventDialogEmptyMode
...
+    public readonly acceptLabel = 'Confirm';
+
+    public startDate: Moment;
+
+    constructor() {
+        super(ClaimSickLeaveDialogModel);
+    }
+
+    public get text(): string {
+        return `Your sick leave has started on ${this.startDate.format(eventDialogTextDateFormat)}`;
```
santee: You sick leave *starts* on

mobile/src/reducers/calendar/event-dialog/event-dialog.model.ts
```diff
@@ -56,6 +64,39 @@ export class EventDialogEmptyModel extends EventDialogModel<EventDialogEmptyMode
 }
 
 export class ClaimSickLeaveDialogModel extends EventDialogModel<ClaimSickLeaveDialogModel> {
+    public readonly title = 'Select date to Start your Sick Leave';
+    public readonly icon = 'sick_leave';
+    public readonly cancelLabel = 'Back';
+    public readonly acceptLabel = 'Confirm';
+
+    public startDate: Moment;
```
santee: should this be publicly accessable for modification?

serjKim: Yes, it should. The idea was to have an ability to clone dialog and change this field in reducers. But _outside_ it is used only as EventDialogModel base type, so no one can access to this field.

mobile/src/reducers/calendar/calendar-events.reducer.ts
serjKim: yes

mobile/src/reducers/calendar/event-dialog/event-dialog.model.ts
santee: why not make a method 'withStartDate(date: Moment) { const instance = this.clone(); instance.startDate = date; return instance; } instead? Otherwise someone will change it.

mobile/src/reducers/calendar/calendar-events.reducer.ts
santee: this does look weird... Ideally, these dialogs should not intersect and calendar should not really know about sick leave events. It should be done somewhere else, and some epic should translate events like CONFIRM_START-DATE-SICK-LEAVE to something like 'NEW-INTERVAL-ADDED'. And it's probably should not happen immediately, I'd wait for HTTP request to finish and THEN send this message

mobile/src/reducers/calendar/calendar.epics.ts
```diff
@@ -1,19 +1,31 @@
...
+    action$.ofType('CONFIRM-CLAIM-SICK-LEAVE')
+        .flatMap(x =>
+            ajaxPost(
+                `${apiUrl}/employees/${x.employee.employeeId}/events`, 
+                x.calendarEvents, 
+                { 'Content-Type': 'application/json' }
+            )
+            .map(obj => deserialize(obj.response, CalendarEvents))
+        )
+        .map(x => loadCalendarEventsFinished([x]))
```
santee: khm, so loadCalendarEventsFinished MERGES new events and existing ones instead of replacing them? I'd suggest to have two different actions.

PR #97, opened 2018-03-30
Added welcome screen, added splash screen, logout, login.

arcelena:


mobile/src/app.tsx
```diff
@@ -10,27 +10,36 @@ import { loadFeeds } from './reducers/feeds/feeds.action';
...
     this.addListener = createReduxBoundAddListener(this.props.reduxNavKey);
   }
 
   public render() {
-    return <RootNavigator navigation={addNavigationHelpers({
-      dispatch: this.props.dispatch as any,
-      state: this.props.nav,
-      addListener: this.addListener
-    })} />;
+    if (this.props.authentication.isAuthenticated === true) {
```
serjKim: without `=== true` )

mobile/src/reducers/auth/auth.reducer.ts
```diff
@@ -0,0 +1,36 @@
...
+import { AuthenticationState } from '../../auth/authentication-state';
+
+export const authEpics$ = combineEpics(
+    pressLogInEpic$ as any,
+    pressLogOutEpic$ as any,
+    listenerAuthStateEpic$ as any
+);
+
+export interface AuthState {
+    isAuthenticated: boolean | null;
```
serjKim: Why do we need `null` state?

mobile/src/app.tsx
arcelena: sure

mobile/src/reducers/auth/auth.reducer.ts
arcelena: It is for splash screen. We need 3 states: 
first one - user is authenticated (isAuthenticated: true)
second one - user is not authenticated (isAuthenticated: false)
third one - app is launching and we have not receive authentication state yet  (isAuthenticated: null).

serjKim: That's actually wrong. `isAuthenticated` should respond only to whether the user is authenticated or not, it shouldn't know about other states such as app is loaded. You should use the other state to track whether app is loaded

serjKim: Quick solution, use a substate:
```javascript
interface AuthInfo {
   isAuthenticated: boolean
}

interface AuthState {
  authInfo: AuthInfo | null;
}

public render() {
     if (!this.props.authentication.authInfo) {
          return <SplashScreen />
     }

     } else if (this.props.authentication.authInfo.isAuthenticated ) {
       return <RootNavigator />;
     } else {
       return <WelcomeScreen />;
     }
   }	   
}
```

mobile/src/reducers/app.reducer.ts
```diff
@@ -44,23 +47,26 @@ const reducers = combineReducers<AppState>({
...
+    const dependencies: DependenciesContainer = { apiClient: new SecuredApiClient(config.apiUrl,  oauthProcess.authenticationState as any ), oauthProcess: oauthProcess };
     const options = { dependencies };
     const epicMiddleware = createEpicMiddleware(rootEpic, options);
 
     const navigationMiddlewareKey = 'root';
 
     const reactNavigationMiddleware = createReactNavigationReduxMiddleware<AppState>(navigationMiddlewareKey, (state) => state.nav);
 
-    return createStore<AppState>(reducers, { navigationMiddlewareKey }, applyMiddleware(epicMiddleware, reactNavigationMiddleware));
+    return createStore<AppState>(reducers, { navigationMiddlewareKey } as AppState, applyMiddleware(epicMiddleware, reactNavigationMiddleware));
```
santee: remove `as AppState` and make authentication optional in interface instead.

mobile/src/reducers/auth/auth.action.ts
```diff
@@ -0,0 +1,13 @@
+export interface PressLogIn {
```
santee: START-LOGIN-PROCESS

mobile/src/reducers/auth/auth.action.ts
```diff
@@ -0,0 +1,13 @@
...
+    type: 'PRESS-LOG-IN';
+}
+
+export const pressLogIn = (): PressLogIn => ({ type: 'PRESS-LOG-IN'});
+
+export interface PressLogOut {
+    type: 'PRESS-LOG-OUT';
```
santee: START-LOGOUT-PROCESS

mobile/src/reducers/auth/auth.epics.ts
```diff
@@ -0,0 +1,26 @@
...
+    action$.ofType('PRESS-LOG-IN')
+        .do(x => {
+            dep.oauthProcess.login();
+        }).ignoreElements();
+
+export const pressLogOutEpic$ = (action$: ActionsObservable<PressLogOut>, state: AppState, dep: DependenciesContainer) =>
+    action$.ofType('PRESS-LOG-OUT')
+        .do(x => {
+            dep.oauthProcess.logout();
+
```
santee: clean empty lines up. Also, you can probably get rid of {} as those are one-liners

mobile/src/reducers/auth/auth.epics.ts
```diff
@@ -0,0 +1,26 @@
...
+        }).ignoreElements();
+
+export const listenerAuthStateEpic$ = (action$: ActionsObservable<any>, state: AppState, dep: DependenciesContainer) =>
+    dep.oauthProcess.authenticationState.map(x => {
+        if (x.isAuthenticated === true) {
+           return pressLogIn();
```
santee: are you sure? I thought we were talking about separate events, why does it dispatches the same commands as button clicks?

mobile/src/reducers/auth/auth.reducer.ts
santee: well, what you've done here is traded `bool` for `AuthInfo` :) Since it's easier to read, I do not mind the change.

mobile/src/reducers/auth/auth.action.ts
```diff
@@ -0,0 +1,25 @@
...
+    type: 'START-LOGOUT-PROCESS';
+}
+
+export const startLogoutProcess = (): StartLogoutProcess => ({ type: 'START-LOGOUT-PROCESS' });
+
+export interface FinishLoginProcess {
+    type: 'FINISH-LOGIN-PROCESS';
+}
+
+export const finishLoginProcess = (): FinishLoginProcess => ({ type: 'FINISH-LOGIN-PROCESS' });
```
santee: -> userLoggedIn

mobile/src/reducers/auth/auth.action.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export interface FinishLogoutProcess {
+    type: 'FINISH-LOGOUT-PROCESS';
+}
+
+export const finishLogoutProcess = (): FinishLogoutProcess => ({ type: 'FINISH-LOGOUT-PROCESS' });
```
santee: -> userLoggedOut

mobile/src/reducers/auth/auth.epics.ts
```diff
@@ -0,0 +1,23 @@
...
+        .do(x => dep.oauthProcess.login())
+        .ignoreElements();
+
+export const startLogoutProcessEpic$ = (action$: ActionsObservable<StartLogoutProcess>, state: AppState, dep: DependenciesContainer) =>
+    action$.ofType('START-LOGOUT-PROCESS')
+        .do(x => dep.oauthProcess.logout())
+        .ignoreElements();
+
+export const listenerAuthStateEpic$ = (action$: ActionsObservable<any>, state: AppState, dep: DependenciesContainer) =>
+    dep.oauthProcess.authenticationState.map(x => {
```
santee: + catch -> userLoggedOut

mobile/src/reducers/auth/auth.reducer.ts
```diff
@@ -0,0 +1,46 @@
...
+    authInfo: AuthInfo | null;
+}
+
+const initState: AuthState = {
+    authInfo: null,
+};
+
+export const authReducer = (state: AuthState = initState, action: AuthActions): AuthState => {
+    switch (action.type) {
+        case 'START-LOGIN-PROCESS':
```
santee: why is start login process here?

mobile/src/reducers/auth/auth.epics.ts
santee: or, better, catch and map to our usual error

PR #100, opened 2018-04-02
Feature/embedded browser for authentication

santee:


mobile/android/app/src/main/java/ru/spb/arcadia/product/assistant/android/AuthenticationSessionModule.java
```diff
@@ -0,0 +1,124 @@
...
+            tryToProcessLogin(intent);
+        }
+    };
+
+
+    public AuthenticationSessionModule(ReactApplicationContext reactContext) {
+        super(reactContext);
+
+        reactContext.addLifecycleEventListener(this);
+        reactContext.addActivityEventListener(this.mEventListener);
```
serjKim: should we unsubscribe?

santee: I *was* looking for a way to do it but I couldn't find anything. The best idea I have is to do it in onHostDestroy

santee: and this is a singleton so we're probably alright...

PR #106, opened 2018-04-06
Feature/process dayoff

serjKim:
Save dayoff


mobile/src/calendar/dayoff-action-button.tsx
```diff
@@ -23,7 +37,18 @@ export class DayoffActionButton extends Component<DayoffActionButtonProps> {
     }
 
     public onDayoffAction = () => {
-        // TODO: do something
+        if (!this.props.interval && this.props.hoursCredit.hours !== 0) {
```
santee: what is hoursCredit? The value that we show on top? Why do you check for it being equal to 0 in that case?

server/Arcadia.Assistant.Calendar/WorkHours/EmployeeWorkHoursActor.cs
```diff
@@ -16,7 +16,7 @@ public class EmployeeWorkHoursActor : CalendarEventsStorageBase
         /// Positive values means that these days must be worked out.
         /// Negative means that these can be taked as days off
         /// </summary>
-        private int hoursCredit = 0;
+        private int hoursCredit = -17;
```
santee: revert

mobile/src/calendar/dayoff-action-button.tsx
serjKim: > what is hoursCredit? The value that we show on top?

it's a model, containg hoursCredit value, and its textual representation - _days + 1/2_. Accordingly, We show textual representation on the top.

santee: why do you check on hours !== 0?

serjKim: I thought, 0 means there is neither dayoff nor workout hours, so disable to process.

PR #109, opened 2018-04-10
Feature/sickleave cancel

serjKim:
Allow cancel sickleave when not approved


serjKim:
Checked. Works fine with the backend

PR #116, opened 2018-04-16
Fix employee id error when feeds loading finished.

arcelena:


mobile/src/reducers/feeds/feed.model.ts
```diff
@@ -13,7 +13,6 @@ export class Feed {
     public messageId: string;
 
     @dataMember()
-    @required()
```
santee: make it nullable instead

PR #128, opened 2018-05-10
 Added opportunity to choose type dayoff/workout

NataliaPonomarevaMM:


serjKim:
Looks good. Add spec for event-dialog.reducer.ts

mobile/src/reducers/calendar/event-dialog/event-dialog.reducer.spec.ts
```diff
@@ -11,10 +13,18 @@ describe('event dialog reducer', () => {
         expect(state.dialogType).toBe(eventTypeDialog);
     });
 
-    it('should return dialog type as null when close', () => {
+    it('should return dialog type as null and dayoff type as default when close', () => {
         const action = closeEventDialog();
         const state = eventDialogReducer(undefined, action);
 
         expect(state.dialogType).toBeNull();
+        expect(state.chosenHoursCreditType).toBe(HoursCreditType.DaysOff);
```
serjKim: You may write another one test ```it('should return dayoff type as default when close')```. Better when one test per one expect.

PR #132, opened 2018-05-16
Added sending email for sick leave

NataliaPonomarevaMM:


server/Arcadia.Assistant.Calendar/SickLeave/SendEmailSickLeaveActor.cs
```diff
@@ -0,0 +1,103 @@
...
+                    new ClientSecrets
+                    {
+                        ClientId = clientId,
+                        ClientSecret = clientSecret,
+                    },
+                    new[] { GmailService.Scope.GmailCompose },
+                    "user",
+                    CancellationToken.None).Result;
+
+                var service = new GmailService(new BaseClientService.Initializer
```
santee: you should not use GmailService, we will use our own SMTP server in production, not Gmail. Gmail supports SMTP, that's what needs to be used (its settings - https://www.lifewire.com/what-are-the-gmail-smtp-settings-1170854 )

server/Arcadia.Assistant.Calendar/SickLeave/SendEmailSickLeaveActor.cs
```diff
@@ -0,0 +1,103 @@
...
+            }
+        }
+
+        private static void SendEmail(SickLeaveIsApproved Event)
+        {
+            var clientId = ConfigurationManager.AppSettings["ClientId"];
+            var clientSecret = ConfigurationManager.AppSettings["ClientSecret"];
+            try
+            {
+                var credential = GoogleWebAuthorizationBroker.AuthorizeAsync(
```
santee: if method returns Task, it should be `await`ed. But it probably doesn't matter since there should be not authorizationbroker in the first place :)

santee: Nevertheless, you should read about async/await - https://docs.microsoft.com/en-us/dotnet/csharp/async
Also, read the documentation here - https://github.com/petabridge/akkadotnet-code-samples/tree/master/PipeTo and this article https://petabridge.com/blog/akkadotnet-async-actors-using-pipeto/ - those two show you have to deal with `Task` in Akka

server/Arcadia.Assistant.Calendar/SickLeave/EmployeeSickLeaveActor.cs
```diff
@@ -10,17 +10,19 @@
 
     public class EmployeeSickLeaveActor : CalendarEventsStorageBase
     {
-        public EmployeeSickLeaveActor(string employeeId)
+        public EmployeeSickLeaveActor(string employeeId, IActorRef sendEmail)
```
santee: -> sendEmailActor.

It is an object/actor, so it should usually be called with noun

server/Arcadia.Assistant.Calendar/SickLeave/SendEmailSickLeaveActor.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        private void SendEmail(SickLeaveIsApproved Event)
+        {
+            var user = mailConfig["SmtpUser"];
+            var pass = mailConfig["SmtpPass"];
+
+            using (var client = new SmtpClient())
+            {
+                var body = "Dear Sir/Madam,\nNotify you, that employee " + Event.UserId +
```
santee: 1) move both body and subject into a configuration file.
2) in all languages, try to avoid string concatenation if possible. Use string.Format or, better, string interpolation

server/Arcadia.Assistant.Calendar/SickLeave/SendEmailSickLeaveActor.cs
```diff
@@ -0,0 +1,68 @@
...
+                    @" went on sick leave for " +
+                    Event.TimeStamp.ToString() + "\nYours faithfully,\nTesting team\n\n";
+                var message = CreateMimeMessage("Approved sick leave", body);
+
+                client.Connect("smtp.gmail.com", 465, SecureSocketOptions.SslOnConnect);
```
santee: host and port should also be configurable

server/Arcadia.Assistant.Calendar/SickLeave/SendEmailSickLeaveActor.cs
santee: If you move it in template, string.Format will be used because the configuration will look as 'aaaaaa from {0} says {1}'

server/Arcadia.Assistant.Calendar/Arcadia.Assistant.Calendar.csproj
```diff
@@ -6,7 +6,10 @@
 
   <ItemGroup>
     <PackageReference Include="Akka" Version="1.3.4" />
+    <PackageReference Include="Akka.DI.Core" Version="1.3.4" />
     <PackageReference Include="Akka.Persistence" Version="1.3.4" />
+    <PackageReference Include="MailKit" Version="2.0.3" />
+    <PackageReference Include="Microsoft.Extensions.Configuration" Version="2.0.0" />
```
santee: Having Configuraiton library linked here is not right, interface should be used instead of IConfigurationSection

server/Arcadia.Assistant.Calendar/SickLeave/EmployeeSickLeaveActor.cs
santee: or, rather, sickListNotifications. Here we don't want to know if it sends an email, or telegram message, or calls api, we just want to tell to notify people.

server/Arcadia.Assistant.Calendar/SickLeave/EmployeeSickLeaveActor.cs
```diff
@@ -10,17 +10,20 @@
...
             : base(employeeId)
         {
             this.PersistenceId = $"employee-sickleaves-{this.EmployeeId}";
+            this.sickLeaveNotifications = sickLeaveNotifications;
         }
 
         public override string PersistenceId { get; }
 
-        public static Props CreateProps(string employeeId)
+        private readonly IActorRef sickLeaveNotifications;
```
santee: a very minor thing, but, in general, people usually keep private readonly fields on very top of the class, then fields, then properties and constructor, then public methods, then private methods. It's easier to read the class declaration this way.

server/Arcadia.Assistant.Calendar/SickLeave/SendEmailSickLeaveActor.cs
```diff
@@ -0,0 +1,59 @@
+﻿using Arcadia.Assistant.Calendar.SickLeave.Events;
+using System;
+using Akka.Actor;
+using Akka.DI.Core;
+using MimeKit;
```
santee: styling. Most of the people prefer using's on top, some of them (including me) - inside namespace. There are arguments for each of these styles, but one thing remains - it should be consistent within project :)

server/Arcadia.Assistant.Calendar/SickLeave/SendEmailSickLeaveActor.cs
```diff
@@ -0,0 +1,59 @@
...
+            }
+        }
+
+        private void SendEmail(SickLeaveIsApproved Event)
+        {
+            using (var client = new SmtpClient())
+            {
+                var message = CreateMimeMessage(String.Format(mailConfig.Body, Event.UserId, Event.TimeStamp.ToString()));
+
+                client.Connect(smtpConfig.Host, smtpConfig.Port, SecureSocketOptions.StartTls);
```
santee: Just for your education.

All I/O methods usually have asynchronous version (that's because they delegate actual work to OS - make a request and wait for response, for instance). The reason is that it doesn't lock a thread, and it can process other requests in the meantime.

Usually, in .NET (or in JS/NodeJS world) you do it using async/await operator. Here, in Akka, it's a bit harder and requires usage of Task.ContinueWith method. Here, however, it's unnecessary, because this code executes very rarely and we can keep it as is.

server/Arcadia.Assistant.Calendar/SickLeave/EmployeeSickLeaveActor.cs
```diff
@@ -179,6 +182,7 @@ private void OnSickleaveApproved(SickLeaveIsApproved message)
             if (this.EventsById.TryGetValue(message.EventId, out var calendarEvent))
             {
                 this.EventsById[message.EventId] = new CalendarEvent(message.EventId, calendarEvent.Type, calendarEvent.Dates, SickLeaveStatuses.Approved, this.EmployeeId);
+                sickLeaveNotifications.Tell(message);
```
santee: Should not be here, but rather in Persist callback. If it's here, this method will be called each time the actor is restored - so it keeps sending emails on system restart

server/Arcadia.Assistant.Calendar/SickLeave/SendEmailSickLeaveActor.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        public static Props GetProps() => Context.DI().Props<SendEmailSickLeaveActor>();
+
+        protected override void OnReceive(object message)
+        {
+            switch (message)
+            {
+                case SickLeaveIsApproved ev:
+                    SendEmail(ev);
```
santee: have you installed Resharper? :) it should complain on lack of `this.` - just a code style issue.

server/Arcadia.Assistant.Calendar/SickLeave/SendEmailSickLeaveActor.cs
```diff
@@ -0,0 +1,58 @@
...
+                    break;
+            }
+        }
+
+        private void SendEmail(SickLeaveIsApproved Event)
```
santee: code style again, parameters should not start with capital letter

server/Arcadia.Assistant.Calendar/SickLeave/SendEmailSickLeaveActor.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            using (var client = new SmtpClient())
+            {
+                var message = CreateMimeMessage(String.Format(mailConfig.Body, Event.UserId, Event.TimeStamp.ToString()));
+
+                client.Connect(smtpConfig.Host, smtpConfig.Port, SecureSocketOptions.StartTls);
```
santee: make start Tls a boolean parameter in the config

server/Arcadia.Assistant.Calendar/SickLeave/SendEmailSickLeaveActor.cs
```diff
@@ -0,0 +1,58 @@
...
+                    SendEmail(ev);
+                    break;
+            }
+        }
+
+        private void SendEmail(SickLeaveIsApproved Event)
+        {
+            using (var client = new SmtpClient())
+            {
+                var message = CreateMimeMessage(String.Format(mailConfig.Body, Event.UserId, Event.TimeStamp.ToString()));
```
santee: add some logging here. Like, 'Sending a sick leave email notification for user {0}'. And one for completion, so it'll help us to see if the message were sent

santee: UserId is pretty useless for accountants :) What they want to see is employee name and its department. So you need to Ask employees actor for some information on the user first (using EmployeeQuery).

server/Arcadia.Assistant.Calendar/SickLeave/SendEmailSickLeaveActor.cs
```diff
@@ -0,0 +1,117 @@
...
+                case SickLeaveIsApproved ev:
+                    if (employeesActor != null)
+                    {
+                        employeesActor.Ask(new EmployeesQuery().WithId(ev.UserId))
+                            .ContinueWith(result => new SickLeaveEmail(ev.TimeStamp, result.Result.AsInstanceOf<EmployeesQuery.Response>()), 
+                                TaskContinuationOptions.AttachedToParent & TaskContinuationOptions.ExecuteSynchronously)
+                            .PipeTo(Self);
+                    }
+                    else
+                        logger.Debug("Employees actor ref is not set in sending email actor");
```
santee: braces. and `this`.  have you installed resharper yet? :)

PR #141, opened 2018-05-31
Fixed handling errors, first load user employee then other requests.

arcelena:


mobile/src/feeds/home-feeds-screen.tsx
```diff
@@ -67,7 +69,9 @@ class HomeFeedsScreenImpl extends React.Component<FeedsScreenProps & FeedScreenD
     }
 
     public componentDidMount() {
-        this.props.fetchNewFeeds();
+        if (this.props.user) {
```
santee: i think not. you need to subscribe to event where you load new users. This way it'll check for user (which might be empty) and won't load anything until you scroll

mobile/src/reducers/feeds/feeds.epics.ts
```diff
@@ -13,9 +13,15 @@ import { Alert } from 'react-native';
 import { retryWhen, map } from 'rxjs/operators';
 import { UnaryFunction } from 'rxjs/interfaces';
 import { handleHttpErrors } from '../errors/errors.epics';
+import { LoadUserEmployeeFinished } from '../user/user.action';
 
 export const pagingPeriodDays = 10;
 
+export const loadUserEmployeeFinishedEpic$ = (action$: ActionsObservable<LoadUserEmployeeFinished>, state: AppState, deps: DependenciesContainer) =>
+    action$.ofType('LOAD-USER-EMPLOYEE-FINISHED')
```
santee: ah, ok, you do it here... but why do you have this call in componentDidMount in this case?

mobile/src/reducers/organization/organization.epics.ts
```diff
@@ -61,4 +61,12 @@ export const loadEmployeesForUserDepartmentEpic$ = (action$: ActionsObservable<L
...
 export const loadEmployeesForUserRoomEpic$ = (action$: ActionsObservable<LoadUserEmployeeFinished>) =>
     action$.ofType('LOAD-USER-EMPLOYEE-FINISHED')
-        .map(x => loadEmployeesForRoom(x.employee.roomNumber));
\ No newline at end of file
+        .map(x => loadEmployeesForRoom(x.employee.roomNumber));
+
+export const loadUserEmployeeFinishedEpic$ = (action$: ActionsObservable<LoadUserEmployeeFinished>, state: AppState, deps: DependenciesContainer) =>
+    action$.ofType('LOAD-USER-EMPLOYEE-FINISHED')
+        .map(x => {
+            console.log('LOAD ROOM');
```
santee: whoops

mobile/src/reducers/errors/errors.epics.ts
```diff
@@ -24,7 +24,7 @@ function retryWhenErrorOccured<T>(isForceLogout: boolean = false): UnaryFunction
         return errors.exhaustMap((e: any) => new Promise((resolve, reject) => {
```
serjKim: typings

PR #183, opened 2018-07-20
Fix Search view height for iphone

serjKim:
simulator screen shot - iphone 6s - 2018-07-20 at 13 23 28


mobile/src/navigation/search-view-styles.tsx
```diff
@@ -27,8 +27,8 @@ export const searchViewStyles = StyleSheet.create({
         fontSize: 18,
     },
     input: {
-        width: '100%',
         color: 'white',
+        backgroundColor: '#4bacca'
```
santee: чубакка...

PR #184, opened 2018-07-20
People/Company: Sort employees

serjKim:


mobile/src/people/employee-comparer.ts
```diff
@@ -1,6 +1,6 @@
 import { Employee } from '../reducers/organization/employee.model';
 
-export function employeesAZSort (first: Employee, second: Employee) {
+export function employeesAZSortComparer (first: Employee, second: Employee) {
```
santee: why SortComparer? :) just Comparer

PR #212, opened 2018-10-16
Add GMG Staff

serjKim:


mobile/src/reducers/people/append-root.ts
```diff
@@ -1,19 +1,24 @@
 import { Department } from '../organization/department.model';
-import { DepartmentIdToNode } from './people.model';
+import { DepartmentIdToNode, DepartmentNode } from './people.model';
 
 export const rootId = '[root]';
+export const gmgId = '[GMG Staff]';
```
santee: NONONONONONO. No hardcode. The difference between 'show people alongside' and 'not' is having people in department (that are not heads of this department). In other words, children of the department should be its child departments + list of employees in this department

PR #240, opened 2018-11-15
Room number is trimmed now

santee:


server/Arcadia.Assistant.CSP/CspEmployeesInfoStorage.cs
```diff
@@ -32,7 +32,7 @@ protected override async Task<LoadAllEmployees.Response> GetAllEmployees()
                             HireDate = x.HiringDate,
                             FireDate = x.FiringDate,
                             MobilePhone = x.MobilePhone,
-                            RoomNumber = x.RoomNumber,
+                            RoomNumber = x.RoomNumber != null ? x.RoomNumber.Trim() : null,
```
sponkrashin: Why not `x.RoomNumber?.Trim()`?

serjKim: expression tree

Issue #231, opened 2018-11-14
Add ping/health endpoint


santee:
should be /health web endpoint, that returns a json with a result of
a) general status
b) server status
c) database/1c maybe?

sponkrashin:
What do you mean by "general status"? Status of API should be determined from application itself. If it cannot get answer from this endpoint then status is not OK.

santee:
general - is 'web' answering or not :) so, basically, by itself it should return 'i'm ok'

sponkrashin:
Added "api/health" endpoint.

KuchaLeni:
Verified

PR #245, opened 2018-11-15
Add ping health endpoint

sponkrashin:
Related to #231


server/Arcadia.Assistant.Health.Abstractions/HealthCheckType.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Arcadia.Assistant.Health.Abstractions
+{
+    public enum HealthCheckType
+    {
+        Server,
+        Check1C,
```
santee: No 1C mentions outside CSPDb. I think the same for Database. Probably CSP implementation should return a dictionary<string, status> - 1c: ok, cspdb: ok

server/Arcadia.Assistant.Health/HealthActor.cs
```diff
@@ -0,0 +1,102 @@
...
+    using Abstractions;
+    using Akka.Actor;
+    using CSP;
+    using CSP.Model;
+    using CSP.Vacations;
+    using Microsoft.EntityFrameworkCore;
+
+    public class HealthActor : UntypedActor, ILogReceive
+    {
+        private readonly Func<ArcadiaCspContext> contextFactory;
```
santee: should not be here. If you need Health implementation, put it in CSP library

server/Arcadia.Assistant.Health/HealthActor.cs
```diff
@@ -0,0 +1,102 @@
...
+            }
+        }
+
+        private async Task<bool> Check1C()
+        {
+            try
+            {
+                var uriBuilder = new UriBuilder(this.bkConfiguration.Url);
+
+                var client = new EnterpriseV8(uriBuilder.Uri)
```
santee: A proper way to do that would be to store Status somewhere in the actor that already talks to database / 1C. When just ask it to return its status.

server/Arcadia.Assistant.Web/Controllers/HealthController.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        [Route("/api/ping/health")]
+        [HttpGet]
+        [ProducesResponseType(typeof(ApplicationHealthModel), StatusCodes.Status200OK)]
+        public async Task<IActionResult> GetApplicationHealth(CancellationToken cancellationToken)
+        {
+            var tasks = new[]
+            {
+                this.healthService.GetIsServerAlive(cancellationToken),
```
santee: what should be happening is that
1. web side requests a health check (with a default timeout * 2)
    -> not given response - return an error.
2. health actor on server side (has CSP implementation)
    -> asks vacationsactor and databaseactor about their status (with defaulttimeout)
    -> return a dictionary.

sponkrashin: Why "* 2"? Why not 3 or 5?

server/Arcadia.Assistant.Health/HealthActor.cs
sponkrashin: Also minus is that currently period of refresh is once in 10 minutes. It is not frequent enough.

server/Arcadia.Assistant.Health.Abstractions/HealthCheckType.cs
sponkrashin: Done.

server/Arcadia.Assistant.Health/HealthActor.cs
sponkrashin: Done.

server/Arcadia.Assistant.Health/HealthActor.cs
sponkrashin: Kept as is - all checks are executed inside health actor. Reasons were listed above.

server/Arcadia.Assistant.Web/Controllers/HealthController.cs
sponkrashin: 1. Web controller makes a request to health service that asks health actor about current server health state.
If actor is not responding - health service returns dictionary with one entry: { server: false }.
If actor is returning result - health service adds entry { server: true } and returns it.
2. On server side health actor makes checks for 1C and database availability by itself and returns dictionary of health state entries. The only remark is that there is no timeout setting in CSP project, it is web specific. So, if either database or 1C check takes a bit more time - server in whole will be considered as not alive.

Is this what you wanted?

server/Arcadia.Assistant.CSP/ArcadiaHealthChecker.cs
```diff
@@ -0,0 +1,97 @@
...
+                    break;
+            }
+        }
+
+        private async Task<bool> Check1C()
+        {
+            try
+            {
+                var uriBuilder = new UriBuilder(this.bkConfiguration.Url);
+                var client = new EnterpriseV8(uriBuilder.Uri)
```
santee: why do that when you can just store success/failure of the last operation?

sponkrashin: > sponkrashin 2 days ago  Member
It is not really proper way, because it violates SRP rule. Actor is starting to do not only what it should do, but also some additional work. But for now let it be so.

> sponkrashin 2 days ago  Member
Also minus is that currently period of refresh is once in 10 minutes. It is not frequent enough.

> sponkrashin 2 days ago  Member
Kept as is - all checks are executed inside health actor. Reasons were listed above.

sponkrashin: Done.

server/Arcadia.Assistant.Web/Controllers/HealthController.cs
sponkrashin: 2. Fixed to get status from existing actor

server/Arcadia.Assistant.CSP/ArcadiaHealthChecker.cs
```diff
@@ -0,0 +1,62 @@
...
+            return new HealthCheckMessageResponse(healthStates);
+        }
+
+        private async Task<HealthState> GetActorHealthState(ActorSelection actor)
+        {
+            try
+            {
+                var result = await actor.Ask<GetHealthCheckStatusMessage.GetHealthCheckStatusResponse>(
+                    GetHealthCheckStatusMessage.Instance,
+                    settings.Timeout);
```
santee: Ok, what I'm afraid of is that if THIS times out, the calling actor (web-part) will also time out, so we won't know which one timed out. That's why I suggested to use default timeout / 2

sponkrashin: ArcadiaHealthChecker actor is called with timeout * 2 value.

sponkrashin: https://github.com/SoftwareCountry/arcadian-assistant/pull/245/files#diff-3f2b01ecf199698605c561a0e4a62ce8R34

server/Arcadia.Assistant.CSP/CspDepartmentsStorage.cs
```diff
@@ -17,12 +18,28 @@ public class CspDepartmentsStorage : DepartmentsStorage
...
         {
             this.contextFactory = contextFactory;
             this.configuration = configuration;
         }
 
+        protected override void OnReceive(object message)
+        {
+            switch (message)
+            {
+                case GetHealthCheckStatusMessage _:
```
santee: maybe put this code in `DepartmentsStorage`? So you won't have a) OnRecieve, b) huge try/catch - just use onFailure callback in `PipeTo` function?

server/Arcadia.Assistant.CSP/CspEmployeesInfoStorage.cs
```diff
@@ -13,40 +13,63 @@
     public class CspEmployeesInfoStorage : EmployeesInfoStorage
     {
         private readonly Func<ArcadiaCspContext> contextFactory;
+        private string lastErrorMessage;
```
santee: same comment I made for DepartmentsStorage

server/Arcadia.Assistant.CSP/CspDepartmentsStorage.cs
santee: oh, nevermind b) part. I think you need to pipe it to self in order to make it work

sponkrashin: Then we obligate dependents of `DepartmentsStorage` to implement this logic even if it uses another health check procedure.

server/Arcadia.Assistant.CSP/ArcadiaHealthChecker.cs
santee: Nevermind, since you have it in a separate variable we can change it

server/Arcadia.Assistant.CSP/CspEmployeesInfoStorage.cs
```diff
@@ -13,40 +13,63 @@
...
-                    })
-                    .ToListAsync();
-                return new LoadAllEmployees.Response(employees);
+                        .ToListAsync();
+                    return new LoadAllEmployees.Response(employees);
+                }
+            }
+            catch (Exception ex)
+            {
+                this.lastErrorMessage = ex.Message;
```
santee: nullify it in try block?

santee: Also, please wrap it using separate method, nesting level is too big

server/Arcadia.Assistant.Health.Abstractions/HealthChecker.cs
```diff
@@ -0,0 +1,25 @@
...
+        protected override void OnReceive(object message)
+        {
+            switch (message)
+            {
+                case HealthCheckMessage _:
+                    this.GetHealthStates().PipeTo(this.Sender);
+                    break;
+
+                default:
+                    Unhandled(message);
```
santee: this.Unhandled :)

server/Arcadia.Assistant.Server.Console/appsettings.json
```diff
@@ -13,6 +13,8 @@
     }
   },
 
+  "TimeoutSeconds": 10,
```
santee: 5 :)

server/Arcadia.Assistant.Web/Authorization/Handlers/UserIsHealthHandler.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Arcadia.Assistant.Web.Authorization.Handlers
+{
+    using System.Threading.Tasks;
+    using Microsoft.AspNetCore.Authorization;
+    using Requirements;
+
+    public class UserIsHealthHandler : AuthorizationHandler<UserIsHealthRequirement>
```
santee: no need for handler, use policy instead to select a scheme

server/Arcadia.Assistant.Web/Authorization/Policies.cs
```diff
@@ -3,5 +3,7 @@
     public static class Policies
     {
         public const string UserIsEmployee = "UserIsEmployee";
+
+        public const string UserIsHealth = "UserIsHealth";
```
santee: use 'ServiceEndpoint' instead. We might use this policy in other places

server/Arcadia.Assistant.Web/Authorization/Requirements/UserIsHealthRequirement.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Arcadia.Assistant.Web.Authorization.Requirements
+{
+    using Microsoft.AspNetCore.Authorization;
+
+    public class UserIsHealthRequirement:IAuthorizationRequirement
```
santee: no need for that

server/Arcadia.Assistant.Web/Controllers/HealthController.cs
```diff
@@ -0,0 +1,46 @@
...
+        [ProducesResponseType(typeof(IEnumerable<ApplicationHealthModelEntry>), StatusCodes.Status503ServiceUnavailable)]
+        public async Task<IActionResult> GetApplicationHealth(CancellationToken cancellationToken)
+        {
+            var healthStateDict = await this.healthService.GetHealthState(cancellationToken);
+
+            var result = healthStateDict
+                .Select(kvp => new ApplicationHealthModelEntry(kvp.Key, kvp.Value.Value, kvp.Value.Details))
+                .ToList();
+
+            if (result.Any(x => !x.StateValue))
```
santee: i think `!result.All(x => x.StateValue)` is more readable, but it's up to you

server/Arcadia.Assistant.Web/Health/HealthService.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public async Task<IDictionary<string, HealthState>> GetHealthState(CancellationToken cancellationToken)
+        {
+            var healthActor = this.actorSystem.ActorSelection(this.pathsBuilder.Get(WellKnownActorPaths.Health));
+
+            try
+            {
+                var healthCheckResponse = await healthActor.Ask<HealthCheckMessageResponse>(
+                    HealthCheckMessage.Instance,
+                    this.timeoutSettings.Timeout * 2,
```
santee: since service timeout can differ from web timeout, you can disregard my magic multiplier :)

server/Arcadia.Assistant.Web/Startup.cs
```diff
@@ -75,18 +75,27 @@ public void ConfigureServices(IServiceCollection services)
...
+                    options =>
+                    {
+                        options.Realm = "Test Realm";
+                        options.EventsType = typeof(AuthenticationEvents);
+                    });
 
             services
-                .AddAuthorization(options => options.AddPolicy(Policies.UserIsEmployee, policy => policy.Requirements.Add(new UserIsEmployeeRequirement())));
+                .AddAuthorization(options => options.AddPolicy(Policies.UserIsEmployee, policy => policy.Requirements.Add(new UserIsEmployeeRequirement())))
+                .AddAuthorization(options => options.AddPolicy(Policies.UserIsHealth, policy => policy.Requirements.Add(new UserIsHealthRequirement())));
```
santee: this AddAuthorization call is not needed, I bet

server/Arcadia.Assistant.Web/Authorization/Handlers/UserIsHealthHandler.cs
sponkrashin: This is WIP, occasionally got into the PR.

server/Arcadia.Assistant.CSP/CspEmployeesInfoStorage.cs
```diff
@@ -32,15 +62,15 @@ protected override async Task<LoadAllEmployees.Response> GetAllEmployees()
                             HireDate = x.HiringDate,
                             FireDate = x.FiringDate,
                             MobilePhone = x.MobilePhone,
-                            RoomNumber = x.RoomNumber != null ? x.RoomNumber.Trim() : null,
+                            RoomNumber = x.RoomNumber,
```
santee: what the heck

sponkrashin: Reverted.

server/Arcadia.Assistant.Web/Health/HealthService.cs
sponkrashin: Done.

server/Arcadia.Assistant.Web/Authorization/Requirements/UserIsHealthRequirement.cs
sponkrashin: Why? It is needed to distinguish authorization handlers.

server/Arcadia.Assistant.Web/Startup.cs
```diff
@@ -75,18 +74,39 @@ public void ConfigureServices(IServiceCollection services)
...
+                        basicOptions.EventsType = typeof(AuthenticationEvents);
+                    });
 
             services
-                .AddAuthorization(options => options.AddPolicy(Policies.UserIsEmployee, policy => policy.Requirements.Add(new UserIsEmployeeRequirement())));
+                .AddAuthorization(options =>
+                {
+                    options.AddPolicy(Policies.UserIsEmployee, policy =>
+                    {
+                        policy.AddAuthenticationSchemes(JwtBearerDefaults.AuthenticationScheme);
```
santee: is this `addauthenticationschemes` required in this policy?

server/Arcadia.Assistant.Web/Startup.cs
```diff
@@ -75,18 +74,39 @@ public void ConfigureServices(IServiceCollection services)
...
+                        policy.Requirements.Add(new UserIsEmployeeRequirement());
+                    });
+
+                    options.AddPolicy(Policies.UserIsServiceUser, policy =>
+                    {
+                        policy.AddAuthenticationSchemes(BasicAuthenticationDefaults.AuthenticationScheme);
+                        policy.Requirements.Add(new ServiceUserRequirement());
```
santee: why requirement is needed?

server/Arcadia.Assistant.Web/Startup.cs
sponkrashin: No. But as I understood, when `Authorize` attribute is used without policy specified, it used first defined policy matched by authorization scheme or without scheme at all. I decided to specify schemes these policies are used with.

server/Arcadia.Assistant.Web/Startup.cs
sponkrashin: To match authorization handler used for specific requirement.

server/Arcadia.Assistant.Web/Startup.cs
```diff
@@ -75,18 +74,39 @@ public void ConfigureServices(IServiceCollection services)
...
+                {
+                    options.AddPolicy(Policies.UserIsEmployee, policy =>
+                    {
+                        policy.AddAuthenticationSchemes(JwtBearerDefaults.AuthenticationScheme);
+                        policy.Requirements.Add(new UserIsEmployeeRequirement());
+                    });
+
+                    options.AddPolicy(Policies.UserIsServiceUser, policy =>
+                    {
+                        policy.AddAuthenticationSchemes(BasicAuthenticationDefaults.AuthenticationScheme);
```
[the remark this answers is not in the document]
sponkrashin: Yes, I experimented with it, but that time I had another errors and it didn't work. I will try again.

server/Arcadia.Assistant.Web/Startup.cs
sponkrashin: Removed.

PR #246, opened 2018-11-16
Fingerprint protection on Android is implemented.

antonm76:
- Fingerprint popup is created according to Material Design guidelines
- Fingerprint popup appears on Splash Screen, proceed with normal login procedure if succeeded; proceed with logout procedure to the Welcome screen if failed.


mobile/src/fingerprint-popup/fingerprint-popup.android.tsx
```diff
@@ -0,0 +1,111 @@
...
+            })
+            .catch((error) => {
+                this.success = false;
+                this.props.onPopupClose();
+            });
+    }
+
+    //----------------------------------------------------------------------------
+    public componentWillUnmount(): void {
+        FingerprintScanner.release();
```
santee: freaking static class... i understand it's not yours :)

mobile/src/splash-screen/splash-screen.tsx
```diff
@@ -1,16 +1,92 @@
...
+}
+
+//============================================================================
+class SplashScreenImpl extends React.Component<SplashScreenDispatchProps, SplashScreenState> {
+
+    //----------------------------------------------------------------------------
+    constructor(props: SplashScreenDispatchProps) {
+        super(props);
 
+        this.onPopupClosed = this.onPopupClosed.bind(this);
```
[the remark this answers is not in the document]
antonm76: Hmm, why is it so? I’m okay with lambdas, but why aren’t methods?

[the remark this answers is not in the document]
santee: When you use simple methods or lambdas it'll compile to
```
function Foo {
//either
this.lambda = () => {}; //autocapture this, create a function instance on each class instance
}
//or
Foo.prototype.method = function() {} // put shared function into prototype and call it every time
```
If you do like you did, it'll have both instance function defined AND prototype method, so in runtime it'll possible to get both these functions and they work differently. You should not notice any problems in your case, but when it comes to inheritance or some runtime magic, it can bit you.
That's from JS point of view. From TS, `this` typing (strict when lambda, not-so-strict when a member) also exist and you sort of deceive them.

mobile/src/fingerprint-popup/fingerprint-popup.android.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+//============================================================================
+export class FingerprintPopupAndroid extends Component<FingerprintPopupProps, FingerprintPopupState> {
+    private success = false;
+
+    //----------------------------------------------------------------------------
+    constructor(props: FingerprintPopupProps) {
+        super(props);
+
+        this.handleAuthenticationAttempt = this.handleAuthenticationAttempt.bind(this);
```
serjKim: use instead of bind:
```typescript
private handleAuthenticationAttempt = (error: Error): void => {
...
}
```

mobile/src/fingerprint-popup/fingerprint-popup.android.tsx
```diff
@@ -0,0 +1,111 @@
...
+    public componentWillUnmount(): void {
+        FingerprintScanner.release();
+    }
+
+    //----------------------------------------------------------------------------
+    public render() {
+        return (
+            <Modal
+                isVisible={this.props.isVisible}
+                style={{ justifyContent: 'flex-end', margin: 0 }}
```
serjKim: move to a separate style

mobile/src/fingerprint-popup/fingerprint-popup.android.tsx
```diff
@@ -0,0 +1,111 @@
...
+                onModalHide={() => { this.props.onPopupHidden(this.success); }}>
```
serjKim: move to a separate method

mobile/src/fingerprint-popup/fingerprint-popup.android.tsx
```diff
@@ -0,0 +1,111 @@
...
+                <View style={FingerprintPopupStyle.view.container}>
+                    <Text style={FingerprintPopupStyle.text.title}>{'Sign in'}</Text>
+                    <View style={[FingerprintPopupStyle.view.fingerImageContainer, this.fingerprintContainerColor()]}>
+                        <Image style={FingerprintPopupStyle.view.fingerImage}
+                               source={require('./fingerprint_white_192x192.png')}/>
```
serjKim: move to somewhere above
const fingerPrintImage = require('./fingerprint_white_192x192.png')

santee: the reason both here and in `onModalHide` is that it returns new reference every `render` call causing view to rerender

mobile/src/splash-screen/splash-screen.tsx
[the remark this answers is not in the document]
antonm76: Addressed in 29b70b8f7989b0d930725c2c79041cf00c4c8d12

mobile/src/fingerprint-popup/fingerprint-popup.android.tsx
antonm76: Addressed in 29b70b8f7989b0d930725c2c79041cf00c4c8d12

mobile/src/fingerprint-popup/fingerprint-popup.android.tsx
antonm76: Addressed in 29b70b8f7989b0d930725c2c79041cf00c4c8d12

mobile/src/fingerprint-popup/fingerprint-popup.android.tsx
antonm76: Addressed in 29b70b8f7989b0d930725c2c79041cf00c4c8d12

mobile/src/fingerprint-popup/fingerprint-popup.android.tsx
antonm76: I see. Addressed in 29b70b8f7989b0d930725c2c79041cf00c4c8d12

mobile/src/app.tsx
```diff
@@ -1,72 +1,70 @@
...
-      return <WelcomeScreen />;
+    public componentDidMount() {
+        BackHandler.addEventListener('hardwareBackPress', this.onBackPress);
     }
-  }
 
-  public componentDidMount() {
-    BackHandler.addEventListener('hardwareBackPress', this.onBackPress );
-  }
+    public componentWillUnmount() {
```
YaroTs: It's deprecated, isn't it?

Issue #250, opened 2018-11-16
Assign a fixed row height on the Feeds view


antonm76:
Currently when application starts and Feeds view is appeared, feed rows grows while additional data are loaded which leads to screen being distorted and rows jumping up and down. Assigning a fixed row height should fix the issue.

KuchaLeni:
Verified

PR #253, opened 2018-11-16
#250: Unnecessary size calculations removed

DesmanLead:
Currently it behaves the same as before, but with less flickering. So the row height is not constant and depends on a device.

Let me now if it requires further optimizations. It worked fast enough on my devices. The only flickering source left is an image component.


santee:
my favorite kind of pull request. @serjKim

PR #256, opened 2018-11-19
Feature/#195 tab bar scroll on double click

YaroTs:
The appropriate tab bar press handler is passed to tab bar navigation options


mobile/src/calendar/calendar.tsx
```diff
@@ -19,7 +20,16 @@ interface CalendarDispatchProps {
     prevCalendarPage: () => void;
 }
 
-export class CalendarImpl extends Component<CalendarProps & CalendarDispatchProps> {
+export class CalendarImpl extends Component<NavigationScreenProps & CalendarProps & CalendarDispatchProps> {
+
+    public componentDidMount() {
+        if (this.props.navigation) {
+            this.props.navigation.setParams({
```
santee: can't you dispatch something instead?

mobile/src/tabbar/tab-bar-on-press-handlers.ts
```diff
@@ -0,0 +1,66 @@
+import {NavigationStateRoute, TabScene} from 'react-navigation';
+import {OnSelectedDayCallback} from '../calendar/calendar-page';
+import moment from 'moment';
+
+export type TabBarOnPress = (options: { scene: TabScene, jumpToIndex: (index: number) => void }) => void;
+
+export const feedsTabBarOnPressHandler: TabBarOnPress = options => {
+    if (options.scene.focused && options.scene.route) {
+
+        let route = options.scene.route as NavigationStateRoute<any>;
```
santee: consts should be your default choice.

mobile/src/calendar/calendar.tsx
YaroTs: Not sure what exactly do you mean. Dispatching events like 'ScrollToTop' from Tabbar?

santee: I don't like that we need to pass navigation property everywhere. react-navigation provides some redux actions for common operations, like navigateTo, or go back. Is it possible to set these params using that method? Also, what happens when this parameter is set once and then we go away from the screen. Is it reset? Because the navigation instance is same.

YaroTs: It seems this parameters cannot be set using redux navigation actions as the tab bar press callback contains only NavigationRoute. 

As the navigation instance is the same, these params will be the same unless somebody manually rewrite them.

santee: in that case, does it make sense to set these parameters once in the root component?

santee: or they differ

santee: seems like they differ, ok, i understand.

mobile/src/reducers/calendar/calendar-events.reducer.ts
```diff
@@ -118,11 +121,13 @@ export const calendarEventsReducer: Reducer<CalendarEventsState> = (state = init
 
             return newState;
         case 'SELECT-CALENDAR-DAY':
+            const initPagesState = resetCalendarPageReducer(state, action);
```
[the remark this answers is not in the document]
YaroTs: That's correct, thank you. Also just noticed that day selection is not needed at all, only page with current month should be displayed.

YaroTs:
Close this pr as react-navigation 3 requires different approach to implement this feature.

Issue #249, opened 2018-11-16
On Android feeds do not load next page when list is scrolled to the bottom


antonm76:
@santee please assign a proper Milestone

antonm76:
Done. develop/dc8c7d8c4669431dd56bf107458f96c0e6d8b6df

KuchaLeni:
Reopened.
(device: OnePlus, android: 9)
Scroll the list of feeds to the bottom and check that loader icon is displayed. 
Result: The next page of feeds doesn't load until the user scrolls up.

antonm76:
@KuchaLeni fixed in 57f6eebf7f5049a4b6be76ad7a2dae69e65bf75d

KuchaLeni:
Verified.

PR #262, opened 2018-11-19
On Android feeds do not load next page when list is scrolled to the bottom #249

antonm76:
1. onEndReachedThreshold has to be >0 to onEndReached to be fired on Android
2. Footer with activity indicator is added

### DIFF
--- a/mobile/src/feeds/home-feeds-screen.tsx
+++ b/mobile/src/feeds/home-feeds-screen.tsx
@@ -61,8 +61,8 @@ class HomeFeedsScreenImpl extends React.Component<FeedsScreenProps & FeedScreenD
         return this.props.feeds.size > 0 ?
             <FlatList
                 style={styles.view}
-                keyExtractor={this.keyExtractor}
-                ItemSeparatorComponent={this.itemSeparator}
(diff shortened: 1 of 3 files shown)


mobile/src/feeds/home-feeds-screen.tsx
```diff
@@ -47,60 +50,82 @@ const mapDispatchToProps = (dispatch: Dispatch<any>): FeedScreenDispatchProps =>
...
+                onEndReachedThreshold={0.2}
+                refreshing={false}
+                onRefresh={this.onRefresh}
+                ListFooterComponent={this.footer}
+            /> :
+            <LoadingView/>;
     }
 
+    //----------------------------------------------------------------------------
+    // noinspection JSMethodCanBeStatic
```
DesmanLead: Why not to actually make it static?

antonm76: I would rather make them lambdas

antonm76: done: 9ade2597c215d48be39585e79cec5fdc7193608e

antonm76:
Done

PR #263, opened 2018-11-19
TSLint semicolon settings are altered

antonm76:
to force class members to be semicoloned, like following:
private endReached = () => {
        if (this.props.user) {
            this.props.fetchOldFeeds();
        }
    };

### DIFF
--- a/mobile/src/app.tsx
+++ b/mobile/src/app.tsx
@@ -58,7 +58,7 @@ export class App extends Component<AppProps> {
 
         dispatch(NavigationActions.back());
         return true;
-    }
+    };
 }
 
 const mapStateToProps = (state: AppState) => ({
--- a/mobile/src/calendar/calendar-action-button.tsx
+++ b/mobile/src/calendar/calendar-action-button.tsx
@@ -63,11 +63,11 @@ export class CalendarActionButton extends Component<CalendarActionButtonProps, C
         this.setState({
             buttonHeight: e.nativeEvent.layout.height
         });
-    }
+    };
 
     private onButtonEditorPress = () => {
         this.props.onPress();
-    }
+    };
 }
 
 export const CalendarActionButtonSeparator = () => <View style={calendarActionsStyles.separator}></View>;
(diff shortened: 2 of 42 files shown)


antonm76:
@serjKim any reasons you are not agree?
why `private flag = true;` requires semicolon while `private foo = () => {}` doesn't?

serjKim:
Because you can't write
```
private flag1 = true private flag2 = true
```
It should be separated by semicolon. Otherwise, after minification it can be broken.

```{}``` is already guaranteed a valid separating code instructions:
```
private a() {} private b() {}
function f1() {} function f2() {}
if (...) {} if (...) {}
```
image
That's why we see, redundant ```;``` after compilation

I don't see any benefits of that proposal

**Comment by antonm76** [comment]
We aren't talking about _methods_, we are talking about _members_.

This TS code
`class A {
    private flag = true;

    private foo = () => {

    };

    private bar = () => {

    };
}`
compiles into following:
`class A {
    constructor() {
        this.flag = true;
        this.foo = () => {
        };
        this.bar = () => {
        };
    }
}`

**Comment by serjKim** [comment]
You are taking about members, containing _body_ {}
As I mentioned, ```{}``` is already separator and can be used without extra ```;```. It's the same if we write that:
```
if () { 
};
while() {
};
```

**Comment by antonm76** [comment]
Hmmm, we have a programming language with a statement terminator, which is a ';'. I prefer do not think whether TS compiles into JS, I prefer to write a code using TS. Also, I prefer uniformity of all kinds, so either we terminate our members declarations with semicolons or we do not.

**Comment by DesmanLead** [comment]
The thing is it is not just a member declaration, it is an assignment statement. And I completely do not understand why the same statement (assignment) should be terminated in one case and should not be terminated in another.

**Comment by serjKim** [comment]
We have to think about JS as well. TS is superset of JS. Unfortunately, It's not a native langauge for us so far.

**Comment by sponkrashin** [comment]
We aren't talking about correctness, we're talking about consistency. And I agree that it looks strange:
```ts
const a = 1;
const b = () => {} 
```

serjKim:
Up to you

sponkrashin:
The main reason to use TSlint in the project is to have well-formatted, consistent code that lesser depends on it's author code-style. Almost all of the TSlint rules doesn't affect performance or any security issues, only code-style. So, in this context, I suppose that this rule can be used.

Issue #191, opened 2018-08-01
'Back' arrow open 'Feeds' tab instead of employee list after department link opening


KuchaLeni:
Steps:
1. Go to the Profile tab.
2. Click on the department link.
3. Click on any employee.
4. Click on 'back' arrow.

Result: Feeds tab is open (see attach).
Expected result: Employees list from user's department is open.
 
[video.zip](https://github.com/SoftwareCountry/arcadian-assistant/files/2249270/video.zip)

KuchaLeni:
The bug is reproduced.

PR #267, opened 2018-11-21
#191: Navigation logic fixed

DesmanLead:
1. Going back on the profile tab will not bring you to the feeds tab anymore;
2. Android back button handling is fixed.

### DIFF
--- a/mobile/src/app.tsx
+++ b/mobile/src/app.tsx
@@ -1,6 +1,5 @@
 import React, { Component } from 'react';
-import { BackHandler, Modal, Text } from 'react-native';
-
+import { BackHandler } from 'react-native';
 import { RootNavigator } from './tabbar/tab-navigator';
 import { AppState } from './reducers/app.reducer';
 import { connect, Dispatch } from 'react-redux';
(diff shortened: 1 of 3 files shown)


mobile/src/app.tsx
```diff
@@ -51,12 +51,13 @@ export class App extends Component<AppProps> {
     }
 
     private onBackPress = () => {
-        const { dispatch, nav } = this.props;
-        if (nav.index === 0) {
+        const { dispatch, nav: currentState } = this.props;
+        const popAction = NavigationActions.pop({ n: 1 });
```
santee: won't it crash if index === 0?

santee: or it's handled with getStateForAction and condition? A question when, if we check for index AND do pop instead of back, will it work? Don't really like getStateForAction call but I can live with it :)

DesmanLead: @santee 
First of all, index here is a tab index (0 == feeds tab). :)
When there is nothing to pop, navigation reducer will just return the same state.

mobile/src/navigation/routes.ts
```diff
@@ -0,0 +1,12 @@
+import { List } from 'immutable';
+
+export const routeComponents = (navigationState: any): List<string> => {
```
santee: Isn't there a type for it?

mobile/src/app.tsx
santee: got it

mobile/src/navigation/routes.ts
DesmanLead: @santee 
There is, but it doesn't describe the full structure.

santee: extend it, maybe? 'any' is just a time bomb.

antonm76:
@DesmanLead are you going to merge it finally?

Issue #224, opened 2018-11-13
Display dayoffs / vacation days on employee details screen


santee:
When user have enough permissions to see it, these values should be displayed

KuchaLeni:
Double click doesn't work on People->Company view

KuchaLeni:
Verified.

PR #275, opened 2018-11-22
Display dayoffs / vacation days on employee details screen #224

Bazt:
Explicit props added to DaysCounters component which are used if provided instead of the current user values from state.

DaysCounters added to the employee details screen


mobile/src/employee-details/employee-details.tsx
```diff
@@ -166,6 +174,27 @@ export class EmployeeDetailsImpl extends Component<EmployeeDetailsProps & Employ
         );
     }
 
+    private renderDaysCounters = (employee: Employee) => {
```
antonm76: Can be a function since doesn't  use 'this' inside

mobile/src/employee-details/employee-details.tsx
```diff
@@ -166,6 +174,27 @@ export class EmployeeDetailsImpl extends Component<EmployeeDetailsProps & Employ
...
+        const calculatedDays = daysConverter.convert(employee.hoursCredit);
+
+        const hoursCreditCounter = new HoursCreditCounter(employee.hoursCredit, calculatedDays.days, calculatedDays.rest);
+
+        const counters: DaysCountersModel = {
+            allVacationDays: allVacationDaysCounter,
+            hoursCredit: hoursCreditCounter,
+        };
+
+        if (employee.vacationDaysLeft !== null || employee.hoursCredit !== null) {
```
antonm76: I would do this condition checking at the very beginning, before creating counters components.

mobile/src/calendar/days-counters/days-counters.tsx
```diff
@@ -11,10 +11,17 @@ interface DaysCountersProps {
...
+
+class DaysCountersImpl extends Component<DaysCountersProps & ExplicitDaysCountersProps> {
 
     public render() {
-        const { daysCounters: { allVacationDays, hoursCredit } } = this.props;
+        const { allVacationDays, hoursCredit } = this.props.explicitCounters ?
+            this.props.explicitCounters :
+            this.props.daysCounters;
 
         if (!allVacationDays && !hoursCredit) {
```
antonm76: Are you sure this condition will never match in case of explicit props passed? Cause it will be infinite in this case.

Bazt:
@santee 
screen shot 2018-11-22 at 15 30 58
screen shot 2018-11-22 at 15 30 27

santee:
I think you might have overdone it, no need to use DaysCounter control, in fact it looks ugly :) We can probably just have same styled lines as Mobile phone and email address, with corresponding icons.

antonm76:
@santee I wonder how many hairs @Bazt snatched when have seen your comment -)

Bazt:
@antonm76 not a single one yet :)

antonm76:
@santee so, do you want it to be merged like it is now, with DaysCounter control, or let @Bazt to refactor it using styled lines?

antonm76:
@santee lets do it in a following way: merge this branch into develop now, so everything @Bazt has done will be kept and let @Bazt to start the new component in a new feature branch. In this case even if it won't be done today we will be able to continue it. @Bazt is a last day in Arcadia today, remember? -)

mobile/src/reducers/calendar/specs/calendar.reducer.spec.ts
```diff
@@ -1,184 +0,0 @@
-import { ConvertHoursCreditToDays } from '../convert-hours-credit-to-days';
-import moment from 'moment';
-import { loadUserEmployeeFinished } from '../../user/user.action';
-import { Employee } from '../../organization/employee.model';
-import { daysCountersReducer } from '../days-counters.reducer';
-
-describe('calendar action', () => {
-    describe('calculate days counters', () => {
```
serjKim: Noooo:sob:

Issue #233, opened 2018-11-14
User preference support


santee:
1. Get push notifications
2. Enable/disable email notifications
~3. Pending requests behavior~

sponkrashin:
Where should user preferences be stored? Database obviously?

sponkrashin:
The 3rd preference is not needed more.

KuchaLeni:
Verified

PR #278, opened 2018-11-22
Feature/user preferences

sponkrashin:
It is for issue #233


server/Arcadia.Assistant.UserPreferences/UserPreferencesActor.cs
```diff
@@ -0,0 +1,63 @@
...
+        public override string PersistenceId => "user-preferences";
+
+        protected override void OnCommand(object message)
+        {
+            switch (message)
+            {
+                case GetUserPreferencesMessage msg:
+                    var userPreferences = this.userPreferencesById.ContainsKey(msg.UserId)
+                        ? this.userPreferencesById[msg.UserId]
+                        : null;
```
santee: what about creating and returning default preferences here? And sending SaveUserPreferencesMessages to Self to store it

server/Arcadia.Assistant.UserPreferences/UserPreferencesActor.cs
```diff
@@ -0,0 +1,63 @@
...
+                    this.Sender.Tell(new GetUserPreferencesMessage.Response(userPreferences));
+                    break;
+
+                case SaveUserPreferencesMessage msg:
+                    try
+                    {
+                        this.Persist(msg, m =>
+                        {
+                            this.SaveUserPreferences(msg.UserId, msg.UserPreferences);
+                            this.Sender.Tell(new SaveUserPreferencesMessage.Success());
```
santee: you sure `this.Sender` will be the same caller? Probably better to have a closure on that?

server/Arcadia.Assistant.UserPreferences/UserPreferencesActor.cs
```diff
@@ -0,0 +1,63 @@
...
+                    var userPreferences = this.userPreferencesById.ContainsKey(msg.UserId)
+                        ? this.userPreferencesById[msg.UserId]
+                        : null;
+                    this.Sender.Tell(new GetUserPreferencesMessage.Response(userPreferences));
+                    break;
+
+                case SaveUserPreferencesMessage msg:
+                    try
+                    {
+                        this.Persist(msg, m =>
```
santee: you never persist a message. Messages define actor contract, it might change (one event can cause multiple events to be fired and so on). What you want to do is to persist events - 'UserChangesPreferences' event, for instance. Also, you can split this event on multiple types - 'UserEnablesEmailNotifications', 'UserEnablesPushNotifications' and handle them differently. For example, when you have devices stored, you might want to remove them from the dictionary of recipients when you handle 'userDisablesPushNotifications' event.

server/Arcadia.Assistant.Web/Controllers/UserPreferencesController.cs
```diff
@@ -0,0 +1,82 @@
...
+                userPreferencesModel,
+                cancellationToken);
+
+            switch (response)
+            {
+                case SaveUserPreferencesMessage.Success _:
+                    return this.NoContent();
+
+                case SaveUserPreferencesMessage.Error error:
+                    return this.BadRequest(error.ErrorMessage);
```
santee: not a bad request, it's 500.

server/Arcadia.Assistant.UserPreferences/UserPreferencesActor.cs
```diff
@@ -0,0 +1,63 @@
...
+            switch (message)
+            {
+                case GetUserPreferencesMessage msg:
+                    var userPreferences = this.userPreferencesById.ContainsKey(msg.UserId)
+                        ? this.userPreferencesById[msg.UserId]
+                        : null;
+                    this.Sender.Tell(new GetUserPreferencesMessage.Response(userPreferences));
+                    break;
+
+                case SaveUserPreferencesMessage msg:
```
santee: you can also compare new preferences to old ones and decide not to store anything

server/Arcadia.Assistant.UserPreferences/UserPreferencesActor.cs
sponkrashin: Why `this` should be changed? Moreover, the same works in `CalendarEventsStorageBase`.

server/Arcadia.Assistant.Web/Controllers/UserPreferencesController.cs
sponkrashin: `CalendarEventsController` returns bad request in such case.

server/Arcadia.Assistant.UserPreferences/UserPreferencesActor.cs
sponkrashin: I use default preferences in user preferences service in Web project. But you're right, it is better to implement it here to allow other actors request for it.

server/Arcadia.Assistant.UserPreferences/UserPreferencesActor.cs
sponkrashin: Done.

server/Arcadia.Assistant.UserPreferences/UserPreferencesActor.cs
sponkrashin: Done.

server/Arcadia.Assistant.UserPreferences/UserPreferencesActor.cs
sponkrashin: Done.

server/Arcadia.Assistant.Web/Controllers/UserPreferencesController.cs
sponkrashin: Fixed.

server/Arcadia.Assistant.UserPreferences/UserPreferencesActor.cs
sponkrashin: Done, but I thought - why do we need to store default value into events? It should always be default if user didn't change them manually. But if we store default preferences to events, they will override if later we change some default behavior (e.g., make notifications disabled by default).

server/Arcadia.Assistant.UserPreferences/UserPreferencesActor.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            if (existingUserPreferences?.DependentDepartmentsPendingActions != message.UserPreferences.DependentDepartmentsPendingActions)
+            {
+                var dependentDepartmentsEvent = new DependentDepartmentsPendingActionsPreferenceChangedEvent(
+                    message.UserId,
+                    message.UserPreferences.DependentDepartmentsPendingActions);
+                this.Persist(dependentDepartmentsEvent, this.OnDependentDepartmentsPendingActionsPreferenceChanged);
+            }
+
+            this.Sender.Tell(new SaveUserPreferencesMessage.Response());
```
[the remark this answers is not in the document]
sponkrashin: It'd make sense if we have several types of responses (Success, NotFound, etc.).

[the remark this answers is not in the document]
santee: but what's the point of waiting for Response? :)

sponkrashin:
You make a request and wait for response. Logic!

пт, 23 нояб. 2018 г., 13:59 Alexander Shevnin notifications@github.com:

PR #283, opened 2018-11-23
#4: Configuration strings extracted into a file

DesmanLead:
ReactNativeConfig is not strictly typed so I believe it is a good idea to leave our wrapper around it.


antonm76:
And, another question, how does it will be possible to set it up on AppCenter, as @santee wants? AFAIK, he wants to set up number of env variables which, in turn, will be used during build in our sources.

DesmanLead:
@antonm76 
1. It is a predefined default name. Any custom name should be specified in an environment variable.
2. package.json was not changed there. Nevertheless, `npm install` changes a lock file.

AppCenter can either override .env or switch between different .env files with an environment variable.

antonm76:
@DesmanLead I see now.

mobile/.env
```diff
@@ -1 +1,6 @@
+appCenterSecretId=09bbbc4f-01bb-4eb9-af8c-516cb4319b5d
```
santee: cut it off pls as well.

mobile/.env.production
```diff
@@ -0,0 +1,6 @@
+appCenterSecretId=prod
+apiUrl=https://arcadia-assistant-dev.arcadialab.ru/api
+oauthRedirectUri=arcadia-assistant://on-login
+oauthClientId=bb342a9b-5cbf-4458-aa1f-88712719774f
```
santee: where did you get this client id? :) and a tenant

Issue #287, opened 2018-11-26
Page refreshing does not update employee information


KuchaLeni:
1) Open the app and go to the Profile view
2) Go to DB and change some employee information (for ex. department 'SDO.822'->'SDO.822.COV')
3) Update the view by swiping down

Result: Info on profile view is not updated
![1](https://user-images.githubusercontent.com/28759011/49027556-7b9a1100-f1b1-11e8-9a26-4dc526bfc693.png)

Expected result: Page refreshing is updates employee information

Post-steps:
1) Go to people view -> department
2) Note that department is updated.
![2](https://user-images.githubusercontent.com/28759011/49027569-85237900-f1b1-11e8-9a43-c40603e0ef16.png)

KuchaLeni:
The same problem on Feeds view. 
@DesmanLead can you fix updating feeds on Feeds view within this issue?

antonm76:
@KuchaLeni what do you mean the same issue? I do not see any department info on the Feeds view. Do you mean department info on the employee view?

KuchaLeni:
> Go to DB and change some employee information 
The problem not only with department (it was just example) but also with email, phone number, etc.

The problem on Feeds view -> new feeds are not uploaded after page refreshing (swipe down). New feed are displayed only after closing and opening the app.

KuchaLeni:
@DesmanLead ok.

PR #290, opened 2018-11-27
#287: shouldComponentUpdate fixed to take department into account

DesmanLead:


mobile/src/profile/home-profile-screen.tsx
```diff
@@ -50,24 +58,39 @@ class ProfileScreenImpl extends Component<ProfileScreenProps & AuthDispatchProps
...
+            }
         }
 
+        const employees = this.props.employees.employeesById;
+        const nextEmployees = nextProps.employees.employeesById;
         if (!employees.equals(nextEmployees)) {
-            const newEmployeesSubset = nextEmployees.filter(employee => {
-                return !employees.has(employee.employeeId);
+            const newEmployeesSubset = nextEmployees.filter(element => {
+                return !employees.has(element.employeeId);
```
antonm76: Shouldn't it be like `if (!employees.has(element.employeeId)) { return true; }`?

DesmanLead: @antonm76 no, this is a filter predicate return.

PR #305, opened 2018-11-29
Select dayoff type dialog switched to radio buttons

DesmanLead:


DesmanLead:
simulator screen shot - iphone se - 2018-11-29 at 20 02 47

mobile/src/common/radio-buttons-group.component.tsx
```diff
@@ -0,0 +1,154 @@
...
+                    ))}
+                </View>
+            </View>
+        );
+    }
+
+    private validate(buttons: RadioButton[]) {
+        let selected = false; // Variable to check if "selected: true" for more than one button.
+        buttons.map(button => {
+            button.color = button.color ? button.color : '#444';
```
serjKim: props shouldn't be mutated

mobile/src/common/radio-buttons-group.component.tsx
```diff
@@ -0,0 +1,154 @@
...
+    public render() {
+        const opacity = this.props.disabled ? 0.2 : 1;
+
+        let layout: ViewStyle = { flexDirection: 'row' };
+        let margin: ViewStyle = { marginLeft: 10 };
+        if (this.props.layout === 'column') {
+            layout = { alignItems: 'center' };
+            margin = { marginTop: 10 };
+        }
+        let textStyle: TextStyle = {
```
serjKim: const textStyle = StyleSheet.flatten([
...
]);
all styles should use StyleSheet.flatten/create, it optimizes the passing ones through the bridge

antonm76: Flattening ruins optimization.
https://github.com/robinpowered/glamorous-native/issues/5
"By using StyleSheet.flatten, we no longer get the performance advantage from registered styles."
https://github.com/facebook/react-native/blob/af949877e67c9ff56089436c7980037a9e332c7c/Libraries/StyleSheet/StyleSheet.js#L201
https://github.com/facebook/react-native/blob/af949877e67c9ff56089436c7980037a9e332c7c/Libraries/StyleSheet/flattenStyle.js#L21

mobile/src/common/radio-buttons-group.component.tsx
antonm76: Not anymore. 3ffda78b754b078f3ac42b00401ac6da16609321

mobile/src/common/radio-buttons-group.component.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+//============================================================================
+export interface RadioButton {
+    selectionIndex: number;
+    label: string;
+
+    labelStyle?: TextStyle;
+    color?: string;
+    disabled?: boolean;
+    layout?: string;
```
serjKim: layout?: 'column' | 'row'

mobile/src/common/radio-buttons-group.component.tsx
```diff
@@ -0,0 +1,175 @@
...
+    size?: number;
+}
+
+//============================================================================
+interface RadioButtonProps extends RadioButton {
+    onPress: (label: string) => void;
+}
+
+//============================================================================
+class RadioButtonComponent extends React.Component<RadioButtonProps> {
```
serjKim: just `RadioButton`

antonm76: there is already interface with this name

Issue #300, opened 2018-11-29
Support new permission: Cancel approved vacations


santee:
Only supervisor should be able to cancel fully APPROVED vacation.
Not yet approved vacations can be canceled by employee himself.
Corresponding information should be returned on a client.

santee:
Only supervisor should be able to cancel fully APPROVED vacation

sponkrashin:
What about other approved events? Workouts, day offs and sick leaves.

santee:
Let's implement the same logic for know

sponkrashin:
Added new employee permission: CancelApprovedCalendarEvents, which is added only to supervisor employees.

KuchaLeni:
Verified.
Supervisor can't cancel vacations with the status 'Processed'. 
Only HR can cancel 'Processed' vacation.

PR #321, opened 2018-12-04
Support new permission cancel approved vacations

sponkrashin:
Issue #300


server/Arcadia.Assistant.Calendar.Abstractions/CalendarEventStatuses.cs
```diff
@@ -73,5 +81,15 @@ public string RejectedForType(string type)
 
             return null;
         }
+
+        public string CancelledByType(string type)
```
santee: ForType, be consistent :)

sponkrashin: Fixed.

PR #325, opened 2018-12-05
Refine vacations pipeline (Part 4. Now surely Final)

sponkrashin:


server/Arcadia.Assistant.Web/Controllers/CalendarEventsController.cs
```diff
@@ -229,24 +205,101 @@ public async Task<IActionResult> Update(string employeeId, string eventId, [From
...
-            string approverId,
-            CancellationToken token)
+        [Route("{eventId}/approvals")]
+        [HttpPost]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status404NotFound)]
+        [ProducesResponseType(StatusCodes.Status409Conflict)]
+        [ProducesResponseType(StatusCodes.Status202Accepted)]
+        public async Task<IActionResult> ApproveEvent(string employeeId, string eventId, CancellationToken token)
```
santee: separate controller.

sponkrashin: Done.

server/Arcadia.Assistant.Web/Controllers/CalendarEventsApprovalsController.cs
```diff
@@ -0,0 +1,166 @@
...
+                return this.Forbid();
+            }
+
+            var requestedEvent = await this.GetCalendarEventOrDefaultAsync(employee.Calendar.CalendarActor, eventId, token);
+            if (requestedEvent == null)
+            {
+                return this.NotFound();
+            }
+
+            if (!requestedEvent.IsPending)
```
santee: why is that? Can't I send approval for already approved request, for example? If I really want to :)

santee: anyway, you don't need to check that here because it throws an error from inside of actor (even if we want that check). And it throws it with a message

sponkrashin: Done.

PR #328, opened 2018-12-05
Feature/#234 user preferences screen

YaroTs:


mobile/src/reducers/user/user-preferences.model.ts
```diff
@@ -0,0 +1,40 @@
+import { dataMember, required } from 'santee-dcts';
+
+export class UserPreferences {
+
+    @dataMember()
+    @required()
+    public emailNotifications: boolean;
```
antonm76: Are you sure you want to use boolean here? And if yes, could you please consider using Optional<boolean>? Cause using 'false' as initial value looks like a quite bad idea.

mobile/src/reducers/user/user.action.ts
```diff
@@ -30,13 +30,44 @@ export interface LoadUserEmployeePermissions extends Action {
...
     ({ type: 'LOAD-USER-EMPLOYEE-PERMISSIONS-FINISHED', permissions });
 
+export interface LoadUserPreferences extends Action {
+    type: 'LOAD-USER-PREFERENCES';
+    userId: string;
+}
+
+export const loadUserPreferences = (userId: string): LoadUserPreferences => ({ type: 'LOAD-USER-PREFERENCES', userId });
+
+export interface UpdateUserPreferences {
```
antonm76: Shouldn't it extends Action?

mobile/src/reducers/user/user.action.ts
```diff
@@ -30,13 +30,44 @@ export interface LoadUserEmployeePermissions extends Action {
...
+}
+
+export const updateUserPreferences = (userId: string, previousPreferences: UserPreferences, newPreferences: UserPreferences): UpdateUserPreferences => ({
+    type: 'UPDATE-USER-PREFERENCES',
+    userId: userId,
+    previousPreferences: previousPreferences,
+    preferences: newPreferences,
+});
+
+export interface LoadUserPreferencesFinished {
```
antonm76: Shouldn't it extends Action?

YaroTs: 6db94d99af30d768fd41fa902218d0b69551520f

mobile/src/reducers/user/user-preferences.model.ts
YaroTs: For me it's better to make UserPreferences nullable in default UserInfoState. Don't you think so?

YaroTs: 68ee907a2dc9b8b30b1a5b4afc4821ada7e05cd9

antonm76: Let be Optional.

mobile/src/user-preferences-screen/settings-view.tsx
```diff
@@ -0,0 +1,33 @@
...
+    }
+});
+
+class SettingsViewImpl extends Component<SettingsDispatchProps> {
+    public render() {
+        return <TouchableOpacity onPress={this.props.onSettingsClicked}>
+            <View style={preferencesStyles.settingsView}>
+                <ApplicationIcon
+                    name={'settings'}
+                    size={Platform.OS === 'ios' ? Dimensions.get('window').width * 0.06 : Dimensions.get('window').width * 0.03}
```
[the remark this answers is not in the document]
YaroTs: 52f547e0835ad66b741c9d95939d54c4dd7fda42

mobile/src/reducers/user/user-info.reducer.ts
```diff
@@ -24,7 +24,7 @@ export const userInfoReducer: Reducer<UserInfoState> = (state = initState, actio
                 ...state,
                 employeeId: action.userEmployeeId
             };
-        case 'LOAD-USER-EMPLOYEE-PERMISSIONS-FINISHED':
+        case 'LOAD-USER-EMPLOYEE-PERMISSIONS-FINISHED': {
```
santee: why block {} ? doesn't make sense in JS context

mobile/src/user-preferences-screen/user-preferences-screen.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+class UserPreferencesScreenImpl extends Component<UserPreferencesScreenProps & UserPreferencesDispatchProps> {
+
+    public static navigationOptions: NavigationScreenConfig<NavigationStackScreenOptions> = {
+        headerTitle: 'Preferences',
+        headerTitleStyle: preferencesStyles.headerTitle,
+    };
+
+    public componentDidMount() {
+        this.loadPreferences();
```
santee: I'm not sure about loading preferences once we open this view. It might be better to load them after user has been loaded and combining them into one model (instead USER_FINISHED + USER_PREFERENCES_FINISHED). Two possible reasons: a) less actions b) some other screens might depend on this model, you don't want this thing to appear in store only after you visited preferences window. The action might stay for 'reload' purpose though.

YaroTs: a4256bedc3b14e82fbd8e9af6d8dbb7083679bc8 Now user preferences are loaded with user employee details just after user loading is finished. Logout happens in case of error.

santee: what's the point in having 'update' and 'load' actions then? There are three /user-preferences/ calls in the code right now, that should not happen

santee: I mean, there are two ways:
1) LOAD(/REFRESH)_USER (calls to user / user-preferences) combined into one model, recall this method on refresh
2) LOAD_USER -> LOAD_USER_FINISHED -> LOAD_USER_PREFERENCES -> LOAD_USER_PREFERENCES_FINISHED, recall this action on refresh.

antonm76: @santee do we aware of preferences being changed before entering Preferences screen?

YaroTs: done 66f53d8a1641ee2faf8130b69e7526f92d6a75fa

mobile/src/reducers/user/user-info.reducer.ts
YaroTs: 2ba853109522b786765bc854b0ffce8496172121

Issue #301, opened 2018-11-29
Show one more screen on dayoff/vacation request submission


santee:
Just informational screen, stating 'Your vacation (dayoff / sick leave) will be shortly reviewed approved by your supervisor". Text can differ for each type, actually

DesmanLead:
@santee is alert fine?
https://facebook.github.io/react-native/docs/alert

antonm76:
@santee is system dialog box enough?

santee:
better to have it as one more screen in the wizard, don't you think?

KuchaLeni:
Reopened.
Informational screen is not displayed for workout and sick leave requests.

sponkrashin:
> it's strange to say e.g. that your sick leave is to be reviewed and approved

Yes, it is correct flow to exclude accidental events that would be processed by HRs otherwise.

DesmanLead:
@sponkrashin the confirmation dialog is there. The task is about one more additional information dialog with just one "ok" button.

sponkrashin:
I just answered to @antonm76's notice that sick leave shouldn't be approved. I don't know anything about your dialog boxes or other stuff :) But as sick leaves should also be approved, dialog box should be there too, imho.

PR #329, opened 2018-12-05
#301: Vacation/Dayoff requests confirmation

DesmanLead:
simulator screen shot - iphone xr - 2018-12-05 at 15 05 26

### DIFF
--- a/mobile/src/calendar/event-dialog/dayoff-requested-dialog.tsx
+++ b/mobile/src/calendar/event-dialog/dayoff-requested-dialog.tsx
@@ -34,6 +34,4 @@ const mapDispatchToProps = (dispatch: Dispatch<Action>): DayoffRequestedDialogDi
     },
 });
 
-const DayoffRequestedDialog = connect(undefined, mapDispatchToProps)(DayoffRequestedDialogImpl);
-
-export default DayoffRequestedDialog;
+export const DayoffRequestedDialog = connect(undefined, mapDispatchToProps)(DayoffRequestedDialogImpl);
--- a/mobile/src/calendar/event-dialog/dayoff-requested-dialog.tsx
+++ b/mobile/src/calendar/event-dialog/dayoff-requested-dialog.tsx
@@ -1,0 +1,39 @@
+/******************************************************************************
+ * Copyright (c) Arcadia, Inc. All rights reserved.
+ ******************************************************************************/
+
+import { Component } from 'react';
+import { EventDialogBase } from './event-dialog-base';
+import React from 'react';
+import { connect } from 'react-redux';
+import { Action, Dispatch } from 'redux';
+import { closeEventDialog } from '../../reducers/calendar/event-dialog/event-dialog.action';
+
+//============================================================================
+interface DayoffRequestedDialogDispatchProps {
+    closeDialog: () => void;
+}
+
+//============================================================================
+class DayoffRequestedDialogImpl extends Component<DayoffRequestedDialogDispatchProps> {
+    public render() {
+        return <EventDialogBase
+            title={'Dayoff requested'}
+            text={'Your dayoff request will be shortly reviewed by your supervisor'}
+            icon={'dayoff'}
+            acceptLabel={'Ok'}
+            onAcceptPress={this.props.closeDialog}
+            onClosePress={this.props.closeDialog} />;
+    }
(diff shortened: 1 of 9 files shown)


mobile/src/calendar/event-dialog/event-dialog-base.tsx
```diff
@@ -69,9 +69,18 @@ export class EventDialogBaseImpl extends Component<EventDialogBaseProps> {
...
+                    <CalendarActionButton title={cancelLabel} onPress={this.cancel}
+                                          style={buttons.cancel as ViewStyle}
+                                          textStyle={buttons.cancelLabel as TextStyle}
+                                          disabled={this.props.inProgress}/>
+                }
+                <CalendarActionButton title={acceptLabel} onPress={this.accept}
+                                      style={buttons.accept as ViewStyle}
+                                      textStyle={buttons.acceptLabel as TextStyle}
+                                      disabled={disableAccept || this.props.inProgress}/>
+                {this.props.inProgress && <ActivityIndicator size={'small'} style={buttons.progressIndicator}/>}
```
antonm76: I would kill for such things:
`{this.props.inProgress && <ActivityIndicator size={'small'} style={buttons.progressIndicator}/>}`

mobile/src/calendar/event-dialog/event-dialog-base.tsx
```diff
@@ -69,9 +69,18 @@ export class EventDialogBaseImpl extends Component<EventDialogBaseProps> {
 
         return (
             <View style={layout.buttons}>
-                <CalendarActionButton title={cancelLabel} onPress={this.cancel} style={buttons.cancel as ViewStyle} textStyle={buttons.cancelLabel as TextStyle} disabled={this.props.inProgress} />
-                <CalendarActionButton title={acceptLabel} onPress={this.accept} style={buttons.accept as ViewStyle} textStyle={buttons.acceptLabel as TextStyle} disabled={disableAccept || this.props.inProgress} />
-                {this.props.inProgress && <ActivityIndicator size={'small'} style={buttons.progressIndicator} />}
+                {
+                    !!cancelLabel &&
```
antonm76: And for such as well... see comment about this.props.inProgress below

mobile/src/reducers/calendar/calendar.epics.ts
```diff
@@ -38,15 +44,16 @@ export const loadCalendarEventsEpic$ = (action$: ActionsObservable<LoadCalendarE
                         .pipe(
                             handleHttpErrors(),
                             map(obj => deserializeArray(obj as any, CalendarEvent)),
-                            map(calendarEvents => new CalendarEvents(calendarEvents))
+                            map(calendarEvents => new CalendarEvents(calendarEvents)),
+                            map(z => {
```
antonm76: Why 'z'???? Why 'x' and hugging 'y'???

mobile/src/calendar/event-dialog/dayoff-requested-dialog.tsx
```diff
@@ -0,0 +1,39 @@
...
+//----------------------------------------------------------------------------
+const mapDispatchToProps = (dispatch: Dispatch<Action>): DayoffRequestedDialogDispatchProps => ({
+    closeDialog: () => {
+        dispatch(closeEventDialog());
+    },
+});
+
+const DayoffRequestedDialog = connect(undefined, mapDispatchToProps)(DayoffRequestedDialogImpl);
+
+export default DayoffRequestedDialog;
```
santee: please avoid default exports (khm, how I forgot to enable it https://palantir.github.io/tslint/rules/no-default-export/).
Three reasons:
1) non-default export allows you to change a variable name in a non-obvious way (otherwise it requires 'as') - and if initially IDE imports in it a proper way, later name changes might not perform depending on personal IDE settings
2) Tree-shaking works way better (the thing that loads only necessary modules) - it's not actually the problem in react-native, I just give you a generic web example
3) In JS/TS world, statements like that are often used by old/deprecated module systems that behave differently, because they actually imported a module, so it's confusing for a reader.

santee: I mean, there is no serious technical reason to use non-default over default, I just explain why this is preferable way :)

PR #330, opened 2018-12-05
Day selection on the Calendar screen is fixed on Android

antonm76:
TapGestureHandler and PanGestureHandler are used. maxDist={50} is set for TapGestureHandler,  minDist={20} is set for PanGestureHandler.


antonm76:
Works on iOS as well.

Issue #304, opened 2018-11-29
Submit approvals/cancellations into CSPDB


santee:
All approvals and cancellations should be inserted into CSPDB.
Vacations / VacationApprovals table. I'll also ask for Cancellations table

KuchaLeni:
Verified

PR #342, opened 2018-12-06
Submit approvals cancellations into cspdb

sponkrashin:
Issue #304

### DIFF
--- a/server/Arcadia.Assistant.CSP/Vacations/ArcadiaVacationRegistry.cs
+++ b/server/Arcadia.Assistant.CSP/Vacations/ArcadiaVacationRegistry.cs
@@ -82,7 +82,7 @@
                     break;
 
                 case CalendarEventCreated msg when msg.Event.Type == CalendarEventTypes.Vacation:
-                    this.UpsertVacation(msg.Event, new string[0])
+                    this.UpsertVacation(msg.Event, null, new string[0])
                         .PipeTo(
                             this.Self,
                             success: () => VacationPersistSuccess.Instance,
@@ -90,7 +90,7 @@
                     break;
 
                 case CalendarEventChanged msg when msg.NewEvent.Type == CalendarEventTypes.Vacation:
-                    this.UpsertVacation(msg.NewEvent, new string[0])
+                    this.UpsertVacation(msg.NewEvent, msg.OldEvent, new string[0])
                         .PipeTo(
                             this.Self,
                             success: () => VacationPersistSuccess.Instance,
@@ -98,7 +98,7 @@
                     break;
 
                 case CalendarEventApprovalsChanged msg when msg.Event.Type == CalendarEventTypes.Vacation:
-                    this.UpsertVacation(msg.Event, msg.Approvals)
+                    this.UpsertVacation(msg.Event, null, msg.Approvals)
                         .PipeTo(
                             this.Self,
                             success: () => VacationPersistSuccess.Instance,
@@ -124,18 +124,15 @@
             }
         }
 
-        private Task<Vacation> UpsertVacation(CalendarEvent vacationCalendarEvent, IEnumerable<string> approvals)
+        private Task<Vacation> UpsertVacation(
+            CalendarEvent @event,
+            CalendarEvent oldEvent,
+            IEnumerable<string> approvals)
         {
-            var vacation = new Vacation
-            {
-                EmployeeId = int.Parse(vacationCalendarEvent.EmployeeId),
-                RaisedAt = DateTimeOffset.Now,
-                Start = vacationCalendarEvent.Dates.StartDate.Date,
-                End = vacationCalendarEvent.Dates.EndDate.Date,
-                Type = this.GetRegularVacationType()
-            };
+            var vacation = this.GetVacationFromCalendarEvent(@event);
+            var oldVacation = this.GetVacationFromCalendarEvent(oldEvent);
 
-            if (vacationCalendarEvent.Status == VacationStatuses.Cancelled)
+            if (@event.Status == VacationStatuses.Cancelled)
             {
                 vacation.CancelledAt = DateTimeOffset.Now;
             }
@@ -153,7 +150,24 @@
                 vacation.VacationApprovals.Add(vacationApproval);
             }
 
-            return this.vacationsPersistenceExecutor.UpsertVacation(vacation);
+            return this.vacationsPersistenceExecutor.UpsertVacation(vacation, oldVacation);
+        }
+
+        private Vacation GetVacationFromCalendarEvent(CalendarEvent @event)
+        {
+            if (@event == null)
+            {
+                return null;
+            }
+
+            return new Vacation
+            {
+                EmployeeId = int.Parse(@event.EmployeeId),
+                RaisedAt = DateTimeOffset.Now,
+                Start = @event.Dates.StartDate.Date,
+                End = @event.Dates.EndDate.Date,
+                Type = this.GetRegularVacationType()
+            };
         }
 
         // ToDo: get it from database
--- a/server/Arcadia.Assistant.CSP/Vacations/ArcadiaVacationRegistry.cs
+++ b/server/Arcadia.Assistant.CSP/Vacations/ArcadiaVacationRegistry.cs
@@ -2,11 +2,16 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading.Tasks;
 
     using Akka.Actor;
     using Akka.Event;
 
+    using Arcadia.Assistant.Calendar.Abstractions;
+    using Arcadia.Assistant.Calendar.Abstractions.EventBus;
     using Arcadia.Assistant.Configuration.Configuration;
+    using Arcadia.Assistant.CSP.Model;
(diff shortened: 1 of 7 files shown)


server/Arcadia.Assistant.CSP/Vacations/ArcadiaVacationRegistry.cs
```diff
@@ -2,27 +2,42 @@
...
     using Arcadia.Assistant.Configuration.Configuration;
+    using Arcadia.Assistant.CSP.Model;
     using Arcadia.Assistant.Organization.Abstractions;
 
     public class ArcadiaVacationRegistry : VacationsRegistry, ILogReceive
     {
         private readonly ILoggingAdapter logger = Context.GetLogger();
 
         private readonly VacationsQueryExecutor vacationsQueryExecutor;
+        private readonly VacationsPersistenceExecutor vacationsPersistenceExecutor;
```
santee: LOL. Persistence executor :) Reading is also a part of Persistence. VacationsDbWrite, VacationsDbReader?

server/Arcadia.Assistant.CSP/Vacations/ArcadiaVacationRegistry.cs
```diff
@@ -66,12 +81,107 @@ protected override void OnReceive(object message)
 
                     break;
 
+                case CalendarEventCreated msg when msg.Event.Type == CalendarEventTypes.Vacation:
```
santee: case CalendarEventCreaten when msg.Event.Type != CalendarEventTypes.Vaction: break? At the very beginning.

server/Arcadia.Assistant.CSP/Vacations/ArcadiaVacationRegistry.cs
```diff
@@ -66,12 +81,107 @@ protected override void OnReceive(object message)
...
+                            success: () => VacationPersistSuccess.Instance,
+                            failure: e => new VacationPersistFailed(e.Message));
+                    break;
+
+                case VacationPersistSuccess _:
+                    this.logger.Debug("Vacation information was updated in the database");
+                    break;
+
+                case VacationPersistFailed err:
+                    this.logger.Warning($"Failed to update vacation information in the database: {err.Message}");
```
santee: can you pass Exception inside?

server/Arcadia.Assistant.CSP/Vacations/VacationsPersistenceExecutor.cs
```diff
@@ -0,0 +1,83 @@
...
+{
+    using System;
+    using System.Linq;
+    using System.Threading.Tasks;
+
+    using Microsoft.EntityFrameworkCore;
+
+    using Arcadia.Assistant.CSP.Model;
+
+    public class VacationsPersistenceExecutor
```
santee: why don't you have a number of methods instead? AddApproval, CancelVacation, RejectVacation? I don't really like the current workflow, you can probably make it much easier?

server/Arcadia.Assistant.CSP/Vacations/ArcadiaVacationRegistry.cs
```diff
@@ -66,12 +81,107 @@ protected override void OnReceive(object message)
...
+
+                case CalendarEventChanged msg when msg.NewEvent.Type == CalendarEventTypes.Vacation:
+                    this.UpsertVacation(msg.NewEvent, msg.OldEvent, new string[0])
+                        .PipeTo(
+                            this.Self,
+                            success: () => VacationPersistSuccess.Instance,
+                            failure: e => new VacationPersistFailed(e.Message));
+                    break;
+
+                case CalendarEventApprovalsChanged msg when msg.Event.Type == CalendarEventTypes.Vacation:
```
santee: Put all this code in a separate child Actor? Why add a hundred LOC in existing one where it doesn't use its features?

sponkrashin: Because you said to add them into existing actor, didn't you? ;)

server/Arcadia.Assistant.CSP/Vacations/ArcadiaVacationRegistry.cs
sponkrashin: `DbWriter` is too common name. As we distinguish db services by their functions, it should be like `VacationsSyncExecutor` (if you like this word so much).

server/Arcadia.Assistant.CSP/Vacations/ArcadiaVacationRegistry.cs
sponkrashin: As you wish, but it doesn't look much prettier.

server/Arcadia.Assistant.CSP/Vacations/ArcadiaVacationRegistry.cs
sponkrashin: Done.

server/Arcadia.Assistant.CSP/Vacations/ArcadiaVacationRegistry.cs
sponkrashin: Done.

server/Arcadia.Assistant.CSP/Vacations/ArcadiaVacationRegistry.cs
sponkrashin: Done.

server/Arcadia.Assistant.CSP/Vacations/ArcadiaVacationRegistry.cs
sponkrashin: Renamed.

server/Arcadia.Assistant.CSP/Vacations/VacationsPersistenceExecutor.cs
sponkrashin: Refactored in attempts to make it more clear.

server/Arcadia.Assistant.CSP/Vacations/VacationsSyncExecutor.cs
```diff
@@ -0,0 +1,185 @@
...
+                var existingVacations = await this.GetVacations(
+                    context,
+                    int.Parse(@event.EmployeeId),
+                    matchInterval.Start,
+                    matchInterval.End);
+
+                if (existingVacations.Count == 0)
+                {
+                    var createdVacation = (await context.Vacations.AddAsync(vacation)).Entity;
+                    existingVacations = new List<Vacation> { createdVacation };
```
santee: not really important, but I'm not happy with creating a new list instead of .Add(createdVacation). I think semantically it's cleaner to not change existing reference

sponkrashin: Sure you're right, especially because it was `HashSet` by default, and becomes `List` after this line.

server/Arcadia.Assistant.CSP/Vacations/VacationsSyncExecutor.cs
```diff
@@ -0,0 +1,185 @@
...
+        private readonly Func<ArcadiaCspContext> contextFactory;
+
+        public VacationsSyncExecutor(Func<ArcadiaCspContext> contextFactory)
+        {
+            this.contextFactory = contextFactory;
+        }
+
+        public async Task<IEnumerable<Vacation>> UpsertVacation(CalendarEvent @event, VacationApprovals approvals, VacationsMatchInterval matchInterval)
+        {
+            var vacation = this.CreateVacationFromCalendarEvent(@event, approvals);
```
santee: please add some NLog's (debug level) with a count of existing vacations, new vacations, ids. Preferably in each case branch so we know what happens.

server/Arcadia.Assistant.Calendar.Abstractions/CalendarEvent.cs
```diff
@@ -16,13 +17,30 @@ public class CalendarEvent
 
         public string EmployeeId { get; }
 
-        public CalendarEvent(string eventId, string type, DatesPeriod dates, string status, string employeeId)
+        public DateTimeOffset CreateDate { get; }
+
+        public DateTimeOffset UpdateDate { get; }
+
+        public string UpdateEmployeeId { get; }
```
santee: I'm not happy about this line. You add some metadata in common class, and only CSP implementation knows how to deal with that. CSP logic depends on exact implementation of this field usage. It violates encapsulation for BOTH pieces. If you need, add multiple fields RejectedBy, CancelledBy, etc.

server/Arcadia.Assistant.Calendar.Abstractions/EventBus/CalendarEventApprovalsChanged.cs
```diff
@@ -1,17 +1,28 @@
...
             this.Approvals = approvals;
         }
 
         public CalendarEvent Event { get; }
 
+        public string LastApproverId { get; }
+
         public IEnumerable<string> Approvals { get; }
+
+        public DateTimeOffset LastApprovalDate { get; }
```
santee: better change it to Approval { ApproverId, TimeStamp}.

PR #343, opened 2018-12-07
Permissions restrict self approve

sponkrashin:


server/Arcadia.Assistant.Organization/DepartmentsSearch.cs
```diff
@@ -77,11 +77,6 @@ private IEnumerable<DepartmentContainer> PostfilterDepartments()
         {
             IEnumerable<DepartmentContainer> result = this.departmentFindings;
 
-            if (this.departmentsQuery.AscendantDepartmentId != null)
```
santee: not used or does not work?

PR #346, opened 2018-12-07
Trying to fix permissions

santee:


server/Arcadia.Assistant.Security/EmployeePermissionsEntry.cs
```diff
@@ -16,7 +16,7 @@ public enum EmployeePermissionsEntry
         RejectCalendarEvents = 1 << 7,
         CompleteSickLeave = 1 << 8,
         ProlongSickLeave = 1 << 9,
-        CancelCalendarEvents = 1 << 10,
+        CancelPendingCalendarEvents = 1 << 10,
```
santee: @antonm76 note of this permission and 'cancelApprovedCalendarEvents' permission below. They are not in use yet.

PR #349, opened 2018-12-10
Login procedure refactoring

YaroTs:
Errors handling modified

### DIFF
--- a/mobile/src/reducers/auth/auth.epics.ts
+++ b/mobile/src/reducers/auth/auth.epics.ts
@@ -26,6 +26,17 @@ function showAlert(message: string, okButtonTitle: string, rejectButtonTitle: st
         }]);
 }
 
+function showErrorMessage(message: string) {
+    Alert.alert(
+        'Error occurred',
+        `${message}`,
+        [{
+            text: 'OK', onPress: () => {
+            }
+        },
+        ]);
+}
+
 export const startLoginProcessEpic$ = (action$: ActionsObservable<StartLoginProcess>, _: StateObservable<AppState>, dep: DependenciesContainer) =>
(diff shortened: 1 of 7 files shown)


mobile/src/root.tsx
```diff
@@ -21,11 +20,6 @@ export class Root extends Component<{}> {
         config.oauth.redirectUri);
 
       this.navigationService = new NavigationService();
-
-      const isIOS = Platform.OS === 'ios';
```
santee: does it still work on Android? I don't remember the code but from the looks of it, something else caused Android app to login, now you call it for both?

YaroTs: Yes, now the login procedure is started in App component for both platforms.

mobile/src/reducers/auth/auth.epics.ts
```diff
@@ -54,16 +65,20 @@ export const listenerAuthStateEpic$ = (action$: ActionsObservable<any>, _: State
     dep.oauthProcess.authenticationState
         .pipe(
             handleHttpErrors(),
-            distinctUntilChanged<AuthenticationState>((x, y) => x.isAuthenticated === y.isAuthenticated),
+            distinctUntilChanged<AuthenticationState>((x, y) => ((x.isAuthenticated === y.isAuthenticated) && x.isAuthenticated)),
```
DesmanLead: Why is this required? Is x == next, y == previous (consider renaming)?

YaroTs: This is needed because if we can't authenticate several times (Authentication state is NotAuthenticated) , we anyway need to show error every time. Without this additional condition no error message will be shown.

PR #351, opened 2018-12-10
Added createdBy, updatedBy, approvedBy and appropriate timestamps to …

sponkrashin:
…all event bus messages.
Issue #304

### DIFF
--- a/server/Arcadia.Assistant.Calendar.Abstractions/EventBus/CalendarEventApprovalsChanged.cs
+++ b/server/Arcadia.Assistant.Calendar.Abstractions/EventBus/CalendarEventApprovalsChanged.cs
@@ -1,38 +1,19 @@
 ﻿namespace Arcadia.Assistant.Calendar.Abstractions.EventBus
 {
-    using System;
     using System.Collections.Generic;
 
     public class CalendarEventApprovalsChanged
     {
         public CalendarEventApprovalsChanged(
             CalendarEvent @event,
-            string lastApprovedBy,
-            DateTimeOffset lastApprovalTimestamp,
-            IEnumerable<string> approvals)
+            IEnumerable<Approval> approvals)
         {
             this.Event = @event;
-            this.LastApproval = new Approval(lastApprovedBy, lastApprovalTimestamp);
             this.Approvals = approvals;
         }
 
         public CalendarEvent Event { get; }
 
-        public Approval LastApproval { get; set; }
-
-        public IEnumerable<string> Approvals { get; }
-
-        public class Approval
-        {
-            public Approval(string approvedBy, DateTimeOffset timestamp)
-            {
-                this.ApprovedBy = approvedBy;
-                this.Timestamp = timestamp;
-            }
-
-            public string ApprovedBy { get; set; }
-
-            public DateTimeOffset Timestamp { get; set; }
-        }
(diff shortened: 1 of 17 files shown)


server/Arcadia.Assistant.Calendar/CalendarEventsStorageBase.cs
```diff
@@ -212,9 +216,11 @@ private void OnCalendarEventNextApproverReceived(string eventId, string nextAppr
                     oldEvent.EmployeeId
                 );
 
-                this.UpdateCalendarEvent(oldEvent, newEvent, ev =>
+                // ToDo: Here we need last approver to set event UserId
```
sponkrashin: Here is the place to pay attention to. This is why I needed those changes that you didn't like with `LastOrDefault`.

server/Arcadia.Assistant.Calendar.Abstractions/EventBus/CalendarEventApprovalsChanged.cs
```diff
@@ -1,17 +1,38 @@
...
             this.Event = @event;
+            this.LastApproval = new Approval(lastApprovedBy, lastApprovalTimestamp);
             this.Approvals = approvals;
         }
 
         public CalendarEvent Event { get; }
 
+        public Approval LastApproval { get; set; }
+
         public IEnumerable<string> Approvals { get; }
```
santee: why don't you have IEnumerable<Approval> Approvals instead. Then you can always find LastApproval by timestamp.

server/Arcadia.Assistant.Calendar/CalendarEventsStorageBase.cs
santee: you don't actually need that. Why do you need that? I mean it's ok to pass null as it was not updated actually.

sponkrashin: But it was updated, indirectly, but was. If someday we have field like "UpdatedBy" in CSP DB - it will fail then.

server/Arcadia.Assistant.Calendar.Abstractions/EventBus/CalendarEventApprovalsChanged.cs
sponkrashin: Done.

server/Arcadia.Assistant.Calendar/CalendarEventsStorageBase.cs
santee: if so eager to use that, why don't you take one from dictionary?

sponkrashin: Because I need your answer. Should the last approver be set as updater of automatically updated calendar event or not?

santee: if that's as easy as this.Approvals[eventId].Approvals.OrderBy(x => x.Timestamp).Select(x => x.UserId).LastOrDefault(), let's do that.
Point is that we don't really need to do that now, but I don't see how it hurts.

sponkrashin: Fixed.

Issue #354, opened 2018-12-11
Send email notifications about calendar event changes


sponkrashin:
Send email notifications about all changes in calendar events (created, cancelled, rejected, approved, user approve granted, completed and so on).

Related to issue #232

sponkrashin:
@santee, which notifications do we need?
I suggest to start with the following:
1) When event is assigned to some approver - send notification to the approver
2) When event status is changed not by owner - send notification to employee of that event
3) When event is changed by owner - send notification to currently assigned approver

santee:
looks good.
+ When new approval granted, send notification to owner.

KuchaLeni:
Verified.
If final approval has been made in the database, only a vacation status change notification will be sent.

PR #358, opened 2018-12-12
Send email notifications about calendar events changes

sponkrashin:
Issue #354


server/Arcadia.Assistant.Server.Console/appsettings.json
```diff
@@ -23,11 +23,33 @@
       "Port": 0,
       "UseTls": true
```
santee: Put 'Enabled' here, since you have it in configuration

sponkrashin: Done.

server/Arcadia.Assistant.Calendar.Notifications/EventAssignedToApproverNotificationActor.cs
```diff
@@ -0,0 +1,125 @@
...
+    using Arcadia.Assistant.Calendar.Abstractions;
+    using Arcadia.Assistant.Calendar.Abstractions.EventBus;
+    using Arcadia.Assistant.Configuration.Configuration;
+    using Arcadia.Assistant.Notifications;
+    using Arcadia.Assistant.Notifications.Email;
+    using Arcadia.Assistant.Organization.Abstractions;
+    using Arcadia.Assistant.Organization.Abstractions.OrganizationRequests;
+    using Arcadia.Assistant.UserPreferences;
+
+    public class EventAssignedToApproverNotificationActor : UntypedActor, ILogReceive
```
santee: that's the thing, it's not really a general notification actor. It has to know all about email/push settings which is not really good. We either need to rename this one, or split it on two (push and emails) with one more layer between them

sponkrashin: We already have common email and common push actors, we don't need more layers.
We can split this on `EventAssignedToApproverEmailNotificationActor` and `EventAssignedToApproverPushNotificationActor` and each actor responsibility will be to wrap and send notification of specific type.
Or we can assume that this concrete actor responsibility is to send notifications of all known types about event assigned to approver.
First is more verbose, but each actor has only one reason to change. Second is less verbose, but actor has several reasons to change. What do you prefer?

sponkrashin: Done.

Issue #359, opened 2018-12-12
Fix how calendar events dates can be changed


sponkrashin:
How it is working now:

Sick Leave
Start date cannot be changed at all. When end date is changed, `SickLeaveProlonged` event is persisted. No current status checks.
 
Work Hours
Dates cannot be changed at all even in requested status.
 
Vacation
No checks at all. Dates can be changed in all statuses. When dates are changed, `VacationDatesAreEdited` event is persisted.

How I suggest to fix it:

User should be able to change dates for all calendar event types in Requested status. In this case `...DatesAreEdited` event should be persisted.
But user should not have ability to change dates if event has at least one user approval.

Sick Leave
When in Approved status it should be possible to change end date with `SickLeaveProlonged` event persisted. In other statuses this ability should be restricted.

Work Hours
Dates cannot be changed in all statuses except Requested.

Vacation
Dates cannot be changed in all statuses except Requested.

sponkrashin:
@santee, any comments on this?

santee:
Sick leave - leave as is

santee:
Daysoff - leave as is

santee:
Vacations can be edited in Requested status only

KuchaLeni:
#511

PR #360, opened 2018-12-13
Fixed how dates changes are handled in different calendar event types.

sponkrashin:
Issue #359


server/Arcadia.Assistant.Calendar/SickLeave/EmployeeSickLeaveActor.cs
```diff
@@ -176,6 +192,14 @@ protected override bool IsStatusTransitionAllowed(string oldCalendarEventStatus,
                 && (newCalendarEventStatus != SickLeaveStatuses.Approved);
         }
 
+        protected override bool IsDatesChangedAllowed(CalendarEvent oldEvent, CalendarEvent newEvent)
+        {
+            var approvals = this.ApprovalsByEvent[oldEvent.EventId];
+
+            return oldEvent.IsPending && approvals.Count == 0 ||
```
santee: well, that needs to be discussed. Put a TODO here, I'm not sure if its right

server/Arcadia.Assistant.Calendar/SickLeave/EmployeeSickLeaveActor.cs
```diff
@@ -188,6 +212,20 @@ private void OnSickLeaveRequest(SickLeaveIsRequested message)
             this.ApprovalsByEvent[message.EventId] = new List<Approval>();
         }
 
+        private void OnSickLeaveDatesEdit(SickLeaveDatesAreEdited message)
```
santee: we do have Prolong. Basically that's what should be allowed. Simple dates changes - no

server/Arcadia.Assistant.Calendar/SickLeave/EmployeeSickLeaveActor.cs
sponkrashin: If later we decide to keep as is - these TODOs will stay forever, nobody will remove them :)

server/Arcadia.Assistant.Calendar/SickLeave/EmployeeSickLeaveActor.cs
sponkrashin: In fact, every employee is just a people, and he can do a mistake. Will he really have to cancel existing event and create new? It is absurdity.

sponkrashin: This is allowed ONLY if calendar event is in requested status. So, mostly it means that event was just created.

server/Arcadia.Assistant.Calendar/SickLeave/EmployeeSickLeaveActor.cs
sponkrashin: We can remark them in separate issue as to be discussed later.

sponkrashin: Reverted.

server/Arcadia.Assistant.Calendar/SickLeave/EmployeeSickLeaveActor.cs
sponkrashin: Reverted.

Issue #232, opened 2018-11-14
Push notifications for approvals


santee:
Send push notifications when new calendar event pending approval / was approved.

We decided to start with notification to the head of department of employee who requests some event. If employee is already head of department, notification should be sent to the head of it's parent department.

KuchaLeni:
Verified

PR #361, opened 2018-12-13
Push notifications support

sponkrashin:
Issue #232

### DIFF
--- a/server/Arcadia.Assistant.Notifications.Push/Events/EmployeeDeviceRemoved.cs
+++ b/server/Arcadia.Assistant.Notifications.Push/Events/EmployeeDeviceRemoved.cs
@@ -1,0 +1,20 @@
+﻿namespace Arcadia.Assistant.Notifications.Push.Events
+{
+    using System;
+
+    public class EmployeeDeviceRemoved
+    {
+        public EmployeeDeviceRemoved(DateTimeOffset timestamp, string employeeId, string deviceId)
+        {
+            this.Timestamp = timestamp;
+            this.EmployeeId = employeeId;
+            this.DeviceId = deviceId;
+        }
+
+        public DateTimeOffset Timestamp { get;  }
+
+        public string EmployeeId { get; }
+
+        public string DeviceId { get;  }
+    }
+}
--- a/server/Arcadia.Assistant.Notifications.Push/PushNotificationsDevicesActor.cs
+++ b/server/Arcadia.Assistant.Notifications.Push/PushNotificationsDevicesActor.cs
@@ -1,0 +1,115 @@
+﻿namespace Arcadia.Assistant.Notifications.Push
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    using Akka.Actor;
+    using Akka.Persistence;
+
+    using Arcadia.Assistant.Notifications.Push.Events;
+
+    public class PushNotificationsDevicesActor : UntypedPersistentActor, ILogReceive
+    {
+        private readonly Dictionary<string, HashSet<string>> devicesByEmployeeId = new Dictionary<string, HashSet<string>>();
+
+        public override string PersistenceId => "push-notifications-devices";
+
+        protected override void OnCommand(object message)
+        {
+            switch (message)
+            {
+                case GetDeviceIds msg:
+                    this.GetDeviceIds(msg);
+                    break;
+
+                case RegisterPushNotificationsDevice msg:
+                    this.RegisterDevice(msg);
+                    break;
+
+                case RemovePushNotificationsDevice msg:
+                    this.RemoveDevice(msg);
+                    break;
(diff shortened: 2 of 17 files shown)


server/Arcadia.Assistant.Notifications.Push/Events/EmployeeDeviceRemoved.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace Arcadia.Assistant.Notifications.Push.Events
+{
+    using System;
+
+    public class EmployeeDeviceRemoved
```
santee: Unregistered? :)

server/Arcadia.Assistant.Notifications.Push/PushNotificationsDevicesActor.cs
```diff
@@ -0,0 +1,115 @@
...
+                case EmployeeDeviceRemoved ev:
+                    this.OnEmployeeDeviceRemoved(ev);
+                    break;
+            }
+        }
+
+        private void GetDeviceIds(GetDeviceIds message)
+        {
+            this.devicesByEmployeeId.TryGetValue(message.EmployeeId, out var deviceIds);
+            this.Sender.Tell(new GetDeviceIds.Success(deviceIds ?? Enumerable.Empty<string>()));
```
santee: better to send a copy of deviceIds for thread-safety reasons

server/Arcadia.Assistant.Notifications.Push/PushNotificationsDevicesActor.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            if (!deviceIds.Contains(@event.DeviceId))
+            {
+                deviceIds.Add(@event.DeviceId);
+            }
+        }
+
+        private void OnEmployeeDeviceRemoved(EmployeeDeviceRemoved @event)
+        {
+            var deviceIds = this.devicesByEmployeeId[@event.EmployeeId];
```
santee: add a check. just in case.

server/Arcadia.Assistant.Notifications.Push/Events/EmployeeDeviceRemoved.cs
sponkrashin: Kept as is.

server/Arcadia.Assistant.Notifications.Push/PushNotificationsDevicesActor.cs
sponkrashin: Done.

server/Arcadia.Assistant.Notifications.Push/PushNotificationsDevicesActor.cs
sponkrashin: Done.

server/Arcadia.Assistant.Web/Controllers/PushNotificationsController.cs
```diff
@@ -0,0 +1,63 @@
...
+        public async Task<IActionResult> RegisterDevice([FromBody]string deviceId, CancellationToken cancellationToken)
+        {
+            var employee = await this.userEmployeeSearch.FindOrDefaultAsync(this.User, cancellationToken);
+
+            await this.pushNotificationsService.RegisterDevice(employee.Metadata.EmployeeId, deviceId, cancellationToken);
+
+            return this.Accepted();
+        }
+
+        [HttpDelete]
```
DesmanLead: I believe the following interface is better:
PUT /api/push/device/{deviceId}
DELETE /api/push/device/{deviceId}
At least because AjaxObservable wrapper doesn't accept body for delete. :)

sponkrashin: Ok, will fix this.

sponkrashin: But I suppose POST fits specification more, because POST to `/device/{deviceId}` creates new device id for current user, not updates existing one. But from the other side, it updates current user devices :)

DesmanLead: AFAIK, PUT and DELETE usually come together. But it's up to you, we can use both.

antonm76: @sponkrashin I would avoid using 'deviceId', in mobile world it has its own semantic. It would be better to have it like 'appCenterInstallId', cause this is exactly what are we going to pass.

antonm76: Well, let it be 'devicePushToken' then, cause on our case appCenterInstallId incapsulates both APNs Device Token and GCM Registration ID/Firebase Token.

PR #365, opened 2018-12-14
Push notifications actor

sponkrashin:
Issue #232

### DIFF
--- a/server/Arcadia.Assistant.Notifications.Push/PushNotificationsActor.cs
+++ b/server/Arcadia.Assistant.Notifications.Push/PushNotificationsActor.cs
@@ -1,8 +1,6 @@
 ﻿namespace Arcadia.Assistant.Notifications.Push
 {
-    using System.Collections.Generic;
     using System.Linq;
-    using System.Net;
     using System.Net.Http;
     using System.Net.Http.Headers;
     using System.Threading.Tasks;
@@ -36,14 +34,6 @@
                     this.SendPushNotification(msg)
                         .PipeTo(
                             this.Self,
-                            success: x => new SendPushNotificationResult(x),
-                            failure: err => new SendPushNotificationResultFinish(err.Message));
-                    break;
-
-                case SendPushNotificationResult msg:
-                    this.HandleSendPushNotificationResults(msg.Responses)
-                        .PipeTo(
-                            this.Self,
                             success: () => SendPushNotificationResultFinish.Instance,
                             failure: err => new SendPushNotificationResultFinish(err.Message));
                     break;
@@ -58,7 +48,7 @@
             }
         }
 
-        private Task<HttpResponseMessage[]> SendPushNotification(PushNotification message)
+        private Task SendPushNotification(PushNotification message)
         {
             this.logger.Debug("Push notification message received");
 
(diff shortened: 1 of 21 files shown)


server/Arcadia.Assistant.Notifications.Push/PushNotificationsActor.cs
```diff
@@ -0,0 +1,141 @@
...
+                .Select(url => this.SendPushNotificationRequest(url, jsonMessage));
+
+            return Task.WhenAll(requests);
+        }
+
+        private async Task HandleSendPushNotificationResults(IEnumerable<HttpResponseMessage> responses)
+        {
+            foreach (var response in responses)
+            {
+                if (response.StatusCode == HttpStatusCode.Accepted)
```
santee: don't check for accepted, check for 2xx status code

server/Arcadia.Assistant.Notifications.Push/PushNotificationsActor.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            var requests = this.pushSettings.ApplicationPushUrls
+                .Select(url => this.SendPushNotificationRequest(url, jsonMessage));
+
+            return Task.WhenAll(requests);
+        }
+
+        private async Task HandleSendPushNotificationResults(IEnumerable<HttpResponseMessage> responses)
+        {
+            foreach (var response in responses)
```
santee: responses should be disposed.

server/Arcadia.Assistant.Notifications.Push/PushNotificationsActor.cs
```diff
@@ -0,0 +1,141 @@
...
+        {
+            this.logger.Debug("Push notification message received");
+
+            var jsonMessage = this.SerializeNotification(message);
+            this.logger.Debug($"Serialized notification message: {jsonMessage}");
+
+            var requests = this.pushSettings.ApplicationPushUrls
+                .Select(url => this.SendPushNotificationRequest(url, jsonMessage));
+
+            return Task.WhenAll(requests);
```
santee: instead of that, you can foreach through returned tasks and Pipe each of them here. When it'll be easier to handle disposables. Also, do you really need 'HandleSendPushNotificationResults'? That comment is just a suggestion.

server/Arcadia.Assistant.Notifications.Push/PushNotificationsActor.cs
```diff
@@ -0,0 +1,141 @@
...
+            var request = new HttpRequestMessage(HttpMethod.Post, applicationPushUrl)
+            {
+                Content = content
+            };
+            request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+            content.Headers.Add("X-API-Token", this.pushSettings.ApiToken);
+
+            using (var httpClient = this.httpClientFactory.CreateClient())
+            {
+                return httpClient.SendAsync(request);
```
santee: i wonder how it works if Http Client might get disposed BEFORE SendAsync finished its job. I assume you just get lucky here because of some pooling. Better to await it here.

server/Arcadia.Assistant.Notifications.Push/PushNotificationsActor.cs
sponkrashin: Done. Recheck, please.

server/Arcadia.Assistant.Notifications.Push/PushNotificationsActor.cs
sponkrashin: Done.

server/Arcadia.Assistant.Notifications.Push/PushNotificationsActor.cs
sponkrashin: Done.

server/Arcadia.Assistant.Notifications.Push/PushNotificationsActor.cs
sponkrashin: Fixed.

Issue #340, opened 2018-12-06
Restrict several vacations with intersected dates


sponkrashin:
Server should respond with Bad Request result if new vacation intersects with existing ones.

KuchaLeni:
Verified

PR #367, opened 2018-12-14
Restrict several calendar events of one type with intersected dates

sponkrashin:
Issue #340

### DIFF
--- a/server/Arcadia.Assistant.Calendar.Abstractions/DatesPeriod.cs
+++ b/server/Arcadia.Assistant.Calendar.Abstractions/DatesPeriod.cs
@@ -64,6 +64,21 @@
             return !Equals(left, right);
         }
 
+        public bool DatesIntersectsWith(DatesPeriod period)
+        {
+            if (period == null)
+            {
+                return false;
+            }
+
+            if (this.EndDate < period.StartDate || period.EndDate < this.StartDate)
+            {
+                return false;
+            }
+
+            return true;
+        }
+
         private static DateTime MinDate(DateTime first, DateTime second)
         {
             return first <= second ? first : second;
--- a/server/Arcadia.Assistant.Calendar.Abstractions/DatesPeriod.cs
+++ b/server/Arcadia.Assistant.Calendar.Abstractions/DatesPeriod.cs
@@ -20,10 +20,10 @@
 
         public DatesPeriod(DateTime startDate, DateTime endDate, int startWorkingHour = 0, int finishWorkingHour = 8)
         {
-            this.StartDate = startDate;
-            this.EndDate = endDate;
-            this.StartWorkingHour = startWorkingHour;
-            this.FinishWorkingHour = finishWorkingHour;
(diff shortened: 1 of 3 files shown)


sponkrashin:
We should, because if anybody calls API request to add vacation with same dates - it will break CSP database.

server/Arcadia.Assistant.Calendar.Abstractions/DatesPeriod.cs
```diff
@@ -20,10 +20,10 @@ public sealed class DatesPeriod
 
         public DatesPeriod(DateTime startDate, DateTime endDate, int startWorkingHour = 0, int finishWorkingHour = 8)
         {
-            this.StartDate = startDate;
-            this.EndDate = endDate;
-            this.StartWorkingHour = startWorkingHour;
-            this.FinishWorkingHour = finishWorkingHour;
+            this.StartDate = MinDate(startDate, endDate);
+            this.EndDate = MaxDate(startDate, endDate);
+            this.StartWorkingHour = Math.Min(startWorkingHour, finishWorkingHour);
```
santee: lol

sponkrashin: It is just to be sure that all first/start params are normalized

santee: I understand that :)

santee: but better to throw an error when created? It's invalid data, after all

santee: Somewhere in controller

server/Arcadia.Assistant.Calendar.Abstractions/DatesPeriod.cs
```diff
@@ -63,5 +63,30 @@ public override int GetHashCode()
...
         }
+
+        public bool DatesIntersectsWith(DatesPeriod period)
+        {
+            if (period == null)
+            {
+                return false;
+            }
+
+            if (this.EndDate < period.StartDate || period.EndDate < this.StartDate)
```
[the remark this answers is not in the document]
santee: nvm, too complicated

server/Arcadia.Assistant.Calendar.Abstractions/DatesPeriod.cs
sponkrashin: It is checked really, it doesn't allow to create event when Finish is more than Start.

santee: what's the purpose of this code then? :) Anyway, better save than sorry?

sponkrashin: Purpose is to not rely on server handles this case :)

PR #368, opened 2018-12-14
Mobile: Push notifications handling

DesmanLead:


mobile/src/notifications/notification-listener.ts
```diff
@@ -0,0 +1,68 @@
...
+        return from(AppCenter.getInstallId());
+    }),
+    map(installId => installIdReceived(installId)),
+);
+
+//----------------------------------------------------------------------------
+const notificationsRegister$ = (action$: ActionsObservable<NotificationAction>, _: StateObservable<AppState>, deps: DependenciesContainer) => action$.pipe(
+    ofType(NotificationActionType.installIdReceived),
+    switchMap(action => {
+        return deps.apiClient.put(`/push/device/${action.installId}`);
```
sponkrashin: This will be `POST` with install id in body.

mobile/src/notifications/notification-listener.ts
```diff
@@ -0,0 +1,68 @@
+/******************************************************************************
```
YaroTs: Why the name of this file is "notification-listener.ts"? All files with epics in this project have names like "something.epics.ts". Consider renaming to "notifications.epics.ts"

mobile/src/notifications/notification-listener.ts
```diff
@@ -0,0 +1,68 @@
...
+
+//----------------------------------------------------------------------------
+const notificationsRegister$ = (action$: ActionsObservable<NotificationAction>, _: StateObservable<AppState>, deps: DependenciesContainer) => action$.pipe(
+    ofType(NotificationActionType.installIdReceived),
+    switchMap(action => {
+        return deps.apiClient.put(`/push/device/${action.installId}`);
+    }),
+    catchError(error => {
+        console.warn(error);
+        return EMPTY;
```
YaroTs: I believe user should be warned that he won't get push notifications in this case.

mobile/src/reducers/auth/auth.epics.ts
```diff
@@ -51,18 +52,28 @@ export const startLoginProcessEpic$ = (action$: ActionsObservable<StartLoginProc
     );
 
 //----------------------------------------------------------------------------
-export const startLogoutProcessEpic$ = (action$: ActionsObservable<StartLogoutProcess>, _: StateObservable<AppState>, dep: DependenciesContainer) =>
+function logout(dependencies: DependenciesContainer, installId?: string) {
+    if (installId) {
+        notificationsUnregister(dependencies, installId).catch(console.warn);
```
YaroTs: As far as I can see the notificationsUnregister contains only the appropriate server call.

If this server call won't succeed, the app will still get push notifications in logout state and try to open profile in logout state, am I right? 
 
If yes is there an app center api to prevent getting notifications in logged out state?

If there is no such api, an additional check in onPushNotificationReceived lister is required IMHO.

mobile/src/notifications/notification-listener.ts
YaroTs: An appropriate alert with only one "Ok" button is enough IMHO, but, of course, it's not a necessary change.

mobile/src/reducers/auth/auth.epics.ts
antonm76: You are absolutely right, but, here is the problem - we cannot, or do not know, how to process silent Push Notifications. In all other cases notification will appear before we could deal with it. So, we sure can check whether user is logged in or compare employeeId but till we do not receive notifications silently it still will be a palliative.

mobile/src/notifications/notification-listener.ts
antonm76: And what be a proper message in this case? Something like "Application wasn't able to register for Push Notifications on the server, Push Notifications will not be received." or something?

YaroTs: Yes, or like "An error occurred ...error description". Push Notifications will not be received. Please contact your service administrators."

Push notifications are insidious, so I want to prevent getting bugs with description like "Push notifications are not received in some cases. Please investigate".

antonm76: Well, makes sense.

mobile/src/reducers/auth/auth.epics.ts
YaroTs: Ok, but are you sure you won't get a crash or an unusual behaviour in case of getting notifications on welcome screen and trying to open profile?

antonm76: Makes sense again. We have to check whether app is authorised and compare employeeId

mobile/src/notifications/notification-listener.ts
DesmanLead: 9aef55b1bc5dec636d8949d4a2c11319e2f551c6

mobile/src/reducers/auth/auth.epics.ts
DesmanLead: 0bf229dbb8bc6bd65937977fdb221da97c6036dc

mobile/src/notifications/notification.epics.ts
```diff
@@ -0,0 +1,75 @@
...
+import AppCenter from 'appcenter';
+import { installIdReceived, NotificationAction, NotificationActionType } from './notification.actions';
+
+//----------------------------------------------------------------------------
+const notificationsHandler$ = (action$: ActionsObservable<UserLoggedIn>, state$: StateObservable<AppState>) =>
+    action$.ofType(AuthActionType.userLoggedIn).pipe(
+        switchMap(() => {
+            const notification$ = new Subject<Action>();
+
+            Push.setListener({
```
santee: Do you do Push.setListener(null) on logout? If not, why?

mobile/src/notifications/notification.epics.ts
```diff
@@ -0,0 +1,75 @@
...
+import { openProfile } from '../navigation/navigation.actions';
+import { AppState, DependenciesContainer } from '../reducers/app.reducer';
+import AppCenter from 'appcenter';
+import { installIdReceived, NotificationAction, NotificationActionType } from './notification.actions';
+
+//----------------------------------------------------------------------------
+const notificationsHandler$ = (action$: ActionsObservable<UserLoggedIn>, state$: StateObservable<AppState>) =>
+    action$.ofType(AuthActionType.userLoggedIn).pipe(
+        switchMap(() => {
+            const notification$ = new Subject<Action>();
```
santee: fyi, nothing restricts you from putting this inner code in a named function or in a class for better readability ;) But'll leave that to you guys.

mobile/src/notifications/notification.epics.ts
santee: ideally, you'd want to do the same when everybody unsubscribes from your newly created subject, but I assume it works here without any issues because setListener removes all previous listeners implicitly

mobile/src/notifications/notification.epics.ts
```diff
@@ -0,0 +1,75 @@
...
+                    if (!notification.customProperties || !notification.customProperties.employeeId) {
+                        return;
+                    }
+
+                    const userInfo = state$.value.userInfo;
+                    if (!userInfo || notification.customProperties.employeeId !== userInfo.employeeId) {
+                        return;
+                    }
+
+                    notification$.next(openProfile());
```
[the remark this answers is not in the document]
antonm76: > Also, I'd suggest to have a class that encapsulates all incoming pushes, transforms them into observable and then subscribe on it. Just a suggestion.

Well, See the point in incapsulating notifications handling into a class! But what is the point of converting them into Observables? Another one hammer?

mobile/src/notifications/notification.epics.ts
DesmanLead: It's better to consider this refactoring when notifications handling logic become more complex. This class interface requirements will be clear then.

mobile/src/notifications/notification.epics.ts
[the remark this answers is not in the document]
DesmanLead: Actually, this is something similar to what I started with. :) But it appeared that with our current notifications handling it is a bit overkill. This refactoring should be done when (and if) our notifications handling become more complex.

mobile/src/notifications/notification.epics.ts
DesmanLead: 41e46c95ce2cac189c0fc6b08e228d14786c8007

PR #369, opened 2018-12-17
Fixed some heisenbugs (hope so!)

santee:


server/Arcadia.Assistant.Web/Controllers/EmployeesController.cs
```diff
@@ -115,19 +116,34 @@ private async Task<EmployeeModel[]> LoadEmployeesAsync(EmployeesQuery query, Can
...
+            return employeeModels;
+        }
+
+        private void FillPhotoUrls(IEnumerable<EmployeeModel> employeeModels)
+        {
+            foreach (var employee in employeeModels)
+            {
+                try
+                {
+                    employee.PhotoUrl = this.Url.Action(nameof(EmployeePhotoController.GetImage), "EmployeePhoto", new { employeeId = employee.EmployeeId }, this.Request.GetUri().Scheme);
```
santee: I assume it failed here because UrlHelper might be not thread-safe.

Issue #341, opened 2018-12-06
Restrict several vacations with intersected dates


sponkrashin:
UI should not allow to select dates for new vacation that intersects with existing vacations (maybe, they should somehow be marked as disabled in calendar).

Related to issue #340

antonm76:
@santee any input on this topic?

KuchaLeni:
Fixed

KuchaLeni:
Fixed

PR #373, opened 2018-12-17
#341: Vacations with intersected dates restricted

DesmanLead:


mobile/src/utils/utils.ts
```diff
@@ -1,10 +1,47 @@
...
 export function getEmployee(state: AppState): Optional<Employee> {
     return (state.organization && state.userInfo && state.userInfo.employeeId) ?
         state.organization.employees.employeesById.get(state.userInfo.employeeId) :
         undefined;
 }
+
+//----------------------------------------------------------------------------
+function getVacationIntervals(intervals: IntervalModel[], exclude: CalendarEvent[] = []): IntervalModel[] {
+    return intervals
+        .filter(interval => interval.calendarEvent.isVacation)
```
sponkrashin: API will restrict all event types if they intersect with other events of the same type. I'm not sure it is correct, we need to ask master @santee.

antonm76: Hmmm. And what?

sponkrashin: What `what`? You application restricts only vacations to have intersected dates, API restricts sick leaves, workouts and day-offs also.

antonm76: "Restrict several vacations with intersected dates", and I believe that was you who created that issue. It also says "UI should not allow to select dates for new vacation that intersects with existing vacations (maybe, they should somehow be marked as disabled in calendar)." So, either you be more specific in issue description or be more specific here.

sponkrashin: Did I blame PR author that he can not read issue description?
I know that issue description is about vacations. I also know that I did server side PR with ALL calendar event types restricted. And moreover, my PR is not merged yet and I don't really know, should we do as you did, or as I did. That's why I need the answer of @santee to judge us.

santee: vacations must not intersect with each other, but they may intersect with other types - e.g. you can chose to work during vacation or create a sick leave while being on vacation.

sponkrashin: Yes, I meant that vacations should not intersect with vacations, sick leaves with sick leaves, and so on.

mobile/src/utils/utils.ts
```diff
@@ -1,10 +1,47 @@
...
+export function isIntersectingAnotherVacation(startDay: DayModel | undefined,
+                                              endDay: DayModel | undefined,
+                                              intervals: ReadOnlyIntervalsModel,
+                                              exclude: CalendarEvent[] = []): boolean {
+    if (!startDay) {
+        return false;
+    }
+
+    if (!endDay) {
+        const intervalsForStartDay = intervals.get(startDay.date);
```
sponkrashin: As I understood, `intervals.get(date)` returns all intervals that starts from specified date. Am I right?

antonm76: Huh, the same question I asked. In fact, intervals are organized so that all days containing in the interval are keys for that interval in the ‘intervals’ structure. So, the code is good.

sponkrashin: So, if there is an event from 01-01-2018 till 5-01-2018, dictionary will contain entries:
```ts
{
    '01-01-2018': [ event ],
    '02-01-2018': [ event ],
    '03-01-2018': [ event ],
    '04-01-2018': [ event ],
    '05-01-2018': [ event ]
}
```
If it is so, it will work.

antonm76: Exactly as you described.

mobile/src/utils/utils.ts
DesmanLead: 22e5f4e63222c2157a937848ec3f4ea77e6367a9

DesmanLead: It's not possible to create intersecting dayoffs now. It's already not possible to create more than one sick leave.

mobile/src/reducers/calendar/specs/open-vacation-event-dialog.epic.spec.ts
```diff
@@ -14,17 +14,6 @@ import { Action } from 'redux';
 import { filter } from 'rxjs/operators';
 
 describe('openEventDialogEpic', () => {
-    describe('when request vacation', () => {
```
santee: other tests, are they still green? If so, let's merge

PR #384, opened 2018-12-19
Feature/appcenter env variables

santee:


antonm76:
Well, I have the only question - does it work?

santee:
Let's merge and find out.

PR #419, opened 2019-01-11
Avoid using of actor context inside async method.

sponkrashin:


server/Arcadia.Assistant.InboxEmail/InboxEmailsNotificator.cs
```diff
@@ -100,15 +107,13 @@ private async Task BroadcastNewEmails()
...
-                    if (this.lastEmailId != null)
-                    {
-                        var emailsEventBus = new EmailsReceivedEventBus(success.Emails.ToList());
-                        Context.System.EventStream.Publish(emailsEventBus);
-                    }
-
                     this.lastEmailId = success.Emails.Max(e => e.UniqueId);
 
-                    break;
+                    var newEmails = this.lastEmailId == null
```
santee: never null?

sponkrashin: Which one? `this.lastEmailId` will be null until any email is loaded. `newEmails` will never be null. It can be, instead of empty enumerable, it may save some processor ticks :)

PR #425, opened 2019-01-14
iOS fingerprint behaviour is refactored to be the same as Android’s one

antonm76:
In previous version of iOS implementation storage itself was fingerprint protected, thus explicit login in the App component did work fine for iOS but lead to wrong behaviour on Android.


santee:
so, quote 'component did work fine for iOS but not for Android'. Title 'iOS behavior is refactored to be the same as Android' is perfect :D

antonm76:
@santee 

> so, quote 'component did work fine for iOS but not for Android'. Title 'iOS behavior is refactored to be the same as Android' is perfect :D

does it mean you are okay with pr? -)

PR #442, opened 2019-01-31
Feature/auth refactoring

santee:
I don't have working Android environment atm, please check if it works there


santee:
I haven't refactored error handling / error messages, but I want to, there is a duplication. Next PR maybe

YaroTs:
Are you going to merge it directly to master?

mobile/src/auth/jwt-token-handler.ts
```diff
@@ -0,0 +1,114 @@
+import { AccessCodeRequest } from './access-code-request';
```
antonm76: It is better to burn this whole module to the ground.

antonm76:
It's too early to merge it. For example, on Android device with connection turned off it works like following:
- Displays Cancel|Try Again dialog for a number of network operations
- Then displays Logout|Try Again dialog (looks like wasn't able to refresh a token)
- When Logout is clicked, turned into red screen but doesn't logout
- After that it looks like all epics are dead but app still doesn't logout for a some period of time
- Then app fall back to Welcome screen...

mobile/src/auth/jwt-token-handler.ts
```diff
@@ -0,0 +1,114 @@
...
+    constructor(
+        public readonly validUntil: moment.Moment,
+        public readonly value: string
+    ) { }
+
+    public get isExpired() {
+        return moment().isSameOrAfter(this.validUntil);
+    }
+
+    public get isExpiredOrAboutTo() {
```
DesmanLead: needsRefresh?

santee: not his responsibility. it won't refresh itself.

DesmanLead: isExpired? It have exactly this meaning for the class user.

santee: `isExpired` says something objective, its about state. 'needsRefresh' is not objective, its subjective and means something for class user only. Why would token need refresh if its state never changes? The only reason it might need refresh if it knows that expired tokens, or tokens that are about to expire are not useful for HTTP requests; moreover, they should be thrown away. Encapsulation violation :P

antonm76:
Why do we need to store jwtToken in redux? If not, then get$ is not needed, subject is not needed etc

mobile/src/auth/jwt-token-handler.ts
DesmanLead: I mean, why not to remove current isExpired and rename isExpiredOrAboutTo to just isExpired?

santee: because it's not actually Expired :) I can remove isExpired as it's not used... I don't know really, I was thinking about this piece myself :)

santee:
@antonm76 it's needed for Avatars. Basically, image component needs to have this value.

PR #464, opened 2019-02-15
Refactor employee vacations

sponkrashin:


server/Arcadia.Assistant.Calendar/Vacations/PersistentEmployeeVacationsRegistryPropsFactory.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Arcadia.Assistant.Calendar.Vacations
+{
+    using Akka.Actor;
+    using Arcadia.Assistant.Calendar.Abstractions;
+
+    public class PersistentEmployeeVacationsRegistryPropsFactory : IEmployeeVacationsRegistryPropsFactory
```
santee: is there a point to register Props factory intefrace, when you can create instances using Props.DI.Create?

sponkrashin: Point is to create concrete instance of vacations registry using employee id right in employee vacations actor.
We can simply DI and create actor in actor system, but then it will not be able to specify concrete employee id and actor will be common for all employee vacations actors.

Issue #465, opened 2019-02-20
List of employees from the department does not load completely 


KuchaLeni:
1. Open the app
2. Log in to the app
3. Click on Employee (not from current user department) on Feeds view
![photo_2019-02-20_19-27-13](https://user-images.githubusercontent.com/28759011/53107283-c2e03580-3545-11e9-8ca9-e4f81dec9d39.jpg)
4. Click on department 
5. Go to the People view -> Company
6. Click on Feed view

**Result:** employee list from step 4 and employee list from step 6 do not match
![photo_2019-02-20_19-27-14](https://user-images.githubusercontent.com/28759011/53107290-c83d8000-3545-11e9-816b-db1e56c9184d.jpg)

KuchaLeni:
Reopened.
Bug is reproduced (the same steps).
![image](https://user-images.githubusercontent.com/28759011/53572771-132a4980-3b7d-11e9-840e-0c96f524f063.png)
![image](https://user-images.githubusercontent.com/28759011/53572784-18879400-3b7d-11e9-905f-a0e167a9c311.png)

DesmanLead:
@KuchaLeni it was actually deployed today. Could you check that again, please?

KuchaLeni:
Now it is fixed.

PR #474, opened 2019-02-25
#465: loading employees for department on department screen

DesmanLead:


mobile/src/navigation/navigation.epics.ts
```diff
@@ -29,6 +35,12 @@ const openDepartment$ = navigateEpic<OpenDepartmentAction>(
     'CurrentDepartment'
 );
 
+//----------------------------------------------------------------------------
+const loadDepartment$ = (action$: ActionsObservable<OpenDepartmentAction>) => action$.pipe(
+    ofType(NavigationActionType.openDepartment),
+    map(x => loadEmployeesForDepartment(x.params.departmentId)),
+);
```
YaroTs: Not sure if navigation.epics is a good place for this one

DesmanLead: Me too. But what is a good place then?

Issue #457, opened 2019-02-12
Change vacations sync direction with CSP DB


sponkrashin:
CSP is considered as source of truth in each cases. Each operation with vacations should be performed using CSP database.
We need to stop using event sourcing for vacations. At application start it should load all information from vacations table and store it into cache. All operations will update this cache and database.

KuchaLeni:
Verified

PR #481, opened 2019-02-26
CSP sync manual db update

sponkrashin:
Issue #457

### DIFF
--- a/server/Arcadia.Assistant.CSP/Vacations/CspEmployeeVacationsRegistry.cs
+++ b/server/Arcadia.Assistant.CSP/Vacations/CspEmployeeVacationsRegistry.cs
@@ -126,17 +126,19 @@
                 case GetCalendarEventApprovals msg:
                     this.GetVacation(msg.Event.EventId)
                         .PipeTo(
+                            this.Self,
                             this.Sender,
-                            success: result =>
-                            {
-                                if (result != null)
-                                {
-                                    return new GetCalendarEventApprovals.SuccessResponse(result.Approvals.ToList());
-                                }
+                            result => new GetCalendarEventApprovalsSuccess(msg.Event.EventId, result),
+                            err => new GetCalendarEventApprovalsError(msg.Event.EventId, err));
+                    break;
 
-                                return new GetCalendarEventApprovals.ErrorResponse($"Vacation with id {msg.Event.EventId} is not found");
-                            },
-                            failure: err => new GetCalendarEventApprovals.ErrorResponse(err.Message));
+                case GetCalendarEventApprovalsSuccess msg:
+                    this.FinishGetCalendarEventApprovals(msg);
+                    break;
+
+                case GetCalendarEventApprovalsError msg:
+                    this.logger.Error(msg.Exception, $"Error occured on get approvals for vacation with id {msg.EventId} from CSP database for employee {this.employeeId}");
+                    this.Sender.Tell(new GetCalendarEventApprovals.ErrorResponse(msg.Exception.ToString()));
                     break;
 
                 case InsertVacation msg:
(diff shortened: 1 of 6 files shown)


server/Arcadia.Assistant.CSP/Vacations/CspEmployeeVacationsRegistry.cs
```diff
@@ -155,17 +196,133 @@ protected override void OnReceive(object message)
...
+            this.UpdateDatabaseVacationsCache(msg.Events.ToList());
+
+            this.ScheduleNextDatabaseRefresh();
         }
 
-        private Task<CalendarEventWithApprovals> GetVacation(string eventId)
+        private void UpdateDatabaseVacationsCache(List<CalendarEventWithApprovals> databaseVacations)
         {
-            return this.vacationsSyncExecutor.GetVacation(this.employeeId, eventId);
+            var createdEvents = new List<CalendarEventWithApprovals>();
```
santee: I'd have this logic in a separate class.
```
var newCache = new DatabaseVacations(list);
var diff = oldCache.diff(newCache)
foreach(diff.updatedEvents)
```

think OOP :P

sponkrashin: Yes, master...

sponkrashin: Done.

server/Arcadia.Assistant.CSP/Vacations/CspEmployeeVacationsRegistry.cs
```diff
@@ -12,66 +12,99 @@
...
+                case Initialize.Success msg:
+                    var eventsById = msg.Events.ToDictionary(x => x.CalendarEvent.EventId);
+                    this.databaseVacationsCache = new DatabaseVacationsCache(eventsById);
+
+                    foreach (var @event in msg.Events.Where(e => this.IsCalendarEventActual(e.CalendarEvent)))
                     {
                         Context.System.EventStream.Publish(new CalendarEventRecoverComplete(@event.CalendarEvent));
                     }
 
+                    this.ScheduleNextDatabaseRefresh();
```
santee: why do this instead of usual scheduling?

server/Arcadia.Assistant.CSP/Vacations/DatabaseVacationsCache.cs
```diff
@@ -0,0 +1,121 @@
...
+            var updatedEvents = new List<CalendarEventWithApprovals>();
+            var approvalsUpdatedEvents = new List<CalendarEventWithApprovals>();
+
+            foreach (var @event in values.Values)
+            {
+                if (!this.cache.ContainsKey(@event.CalendarEvent.EventId))
+                {
+                    if (@event.CalendarEvent.Status == VacationStatuses.Requested)
+                    {
+                        if (!@event.Approvals.Any())
```
santee: can it be BOTH newly created but with some approvals added?

server/Arcadia.Assistant.CSP/Vacations/CspEmployeeVacationsRegistry.cs
sponkrashin: Because when all events are requested I also update cache. Thereafter we need next update in 10 minutes after current moment, not after previous update.

server/Arcadia.Assistant.CSP/Vacations/DatabaseVacationsCache.cs
```diff
@@ -0,0 +1,121 @@
...
+                        }
+                    }
+                    else
+                    {
+                        updatedEvents.Add(@event);
+                    }
+                }
+                else
+                {
+                    var cacheEvent = this.cache[@event.CalendarEvent.EventId];
```
santee: TryGetValue(out var cacheEvent) instead of ContainsKey

server/Arcadia.Assistant.CSP/Vacations/DatabaseVacationsCache.cs
sponkrashin: If event was "created", it will also flow all the flow needed to get next approval and be added to pending actions.

server/Arcadia.Assistant.CSP/Vacations/DatabaseVacationsCache.cs
sponkrashin: Ok, but what's real profit?

sponkrashin: Done.

server/Arcadia.Assistant.CSP/Vacations/CspEmployeeVacationsRegistry.cs
santee: just send a message to self? When all events are requested you need to update it first? So you should a) send a message to update, b) forward message on get events. no?

sponkrashin: 1) ScheduleNextDatabaseRefresh is needed to prevent getting data from the database too often. If 9 minutes have already passed since previous update and all events are requested - there is no need to update them again after 1 more minute. That's why I cancel previous scheduled updated and schedule new.

2)  Why need to send one more message, if we can just update current cache in "get events" message?

santee: ok.

Issue #482, opened 2019-02-27
Processed vacation does not have status 


KuchaLeni:
Blank line is displayed instead of event status (for ex. 'has coming vacation')
![image](https://user-images.githubusercontent.com/28759011/53477882-b187b380-3a86-11e9-8cc9-2de8a1a3b9e3.png)

antonm76:
@KuchaLeni it would be great to have a mobile label set on this issue -)

santee:
right... in that case, approved vacations should be displayed the same as requested (including style and text).
'has coming vacation' should be displayed for processed events.

antonm76:
@santee Node, already implemented it this way -)

sponkrashin:
Maybe for Approved it should be like "Sent to accounting department"?

santee:
no.

sponkrashin:
@antonm76, as I can see in 9a521de298d49e59ecc71431b73f4ec84dc32221 approved and processed vacations has the same status for now. But as I understood, @santee meant that Approved should become as Requested and Processed as Approved.

antonm76:
@sponkrashin yes, and it looks that @santee just misspelled, he also said "
'has coming vacation' should be displayed for processed events" and this corresponds to Approved. And as far as I can see, Processed is more like Approved than Requested, cause it happens when request it Approved.

sponkrashin:
I think it is not misspell.

> 'has coming ...' should be displayed for processed

> approved vacations should be displayed the same as requested

All is clear, as for me :)

antonm76:
you are right

KuchaLeni:
Fixed

PR #499, opened 2019-02-28
#482 Processed vacation does not have status

antonm76:
“in that case, approved vacations should be displayed the same as requested (including style and text).
'has coming vacation' should be displayed for processed events.” (C) Alex Shevnin

### DIFF
--- a/mobile/src/employee-details/employee-details-events-list.tsx
+++ b/mobile/src/employee-details/employee-details-events-list.tsx
@@ -4,7 +4,7 @@ import { Dimensions, FlatList, ListRenderItemInfo, View, ViewStyle } from 'react
 import { StyledText } from '../override/styled-text';
 import { Avatar } from '../people/avatar';
 import { layoutStylesForEmployeeDetailsScreen } from './styles';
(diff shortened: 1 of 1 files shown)


mobile/src/employee-details/employee-details-events-list.tsx
```diff
@@ -122,15 +122,39 @@ export class EmployeeDetailsEventsList extends Component<EmployeeDetailsEventsLi
...
+
+    //----------------------------------------------------------------------------
+    // noinspection JSMethodCanBeStatic
+    private preprocessStatus(event: CalendarEvent): CalendarEventStatus {
+        if (!event.isVacation) {
+            return event.status;
+        }
+
+        switch (event.status) {
+            case CalendarEventStatus.Requested:
```
sponkrashin: You perceived Alex's words too literally, as for me, but ok :)

antonm76: Well, this leaves the descriptionStatus function intact, I didn't want to deal with if's conditions there.

Issue #549, opened 2019-03-22
Add vacation cancel reason to push notification


sponkrashin:
If vacation cancel reason is set in CSP database, it should be added to push notification when vacation is cancelled.

sponkrashin:
If vacation cancel reason is set in CSP database, it should be added to push notification when vacation is cancelled.

KuchaLeni:
Reopened.
![image](https://user-images.githubusercontent.com/28759011/55564722-3064b180-5701-11e9-99cf-0ad15d0c294d.png)
![image](https://user-images.githubusercontent.com/28759011/55564767-4d00e980-5701-11e9-9fcc-2d4990439952.png)

sponkrashin:
Deployment problems. Now it works.

KuchaLeni:
Fixed.
![image](https://user-images.githubusercontent.com/28759011/55872354-73a5a100-5b95-11e9-82ca-a1c3409fbf54.png)
![image](https://user-images.githubusercontent.com/28759011/55872410-93d56000-5b95-11e9-9c21-a1d771f0d55a.png)

PR #550, opened 2019-03-24
Add vacation cancel reason

sponkrashin:
Issue #549

### DIFF
--- a/server/Arcadia.Assistant.Calendar.Abstractions/CalendarEvent.cs
+++ b/server/Arcadia.Assistant.Calendar.Abstractions/CalendarEvent.cs
@@ -1,5 +1,6 @@
 ﻿namespace Arcadia.Assistant.Calendar.Abstractions
 {
+    using System.Collections.Generic;
     using System.Linq;
 
     public class CalendarEvent
@@ -16,18 +17,22 @@
 
         public string EmployeeId { get; }
 
+        public IReadOnlyDictionary<string, string> AdditionalData { get; }
+
         public CalendarEvent(
             string eventId,
             string type,
             DatesPeriod dates,
             string status,
-            string employeeId)
+            string employeeId,
+            IReadOnlyDictionary<string, string> additionalData = null)
         {
             this.EventId = eventId;
             this.Dates = dates;
             this.Status = status;
             this.Type = type;
             this.EmployeeId = employeeId;
+            this.AdditionalData = additionalData;
             this.IsPending = new CalendarEventStatuses().PendingForType(type).Contains(status);
         }
     }
--- a/server/Arcadia.Assistant.Calendar.Abstractions/CalendarEvent.cs
+++ b/server/Arcadia.Assistant.Calendar.Abstractions/CalendarEvent.cs
@@ -32,7 +32,7 @@
             this.Status = status;
             this.Type = type;
             this.EmployeeId = employeeId;
-            this.AdditionalData = additionalData;
+            this.AdditionalData = additionalData ?? new Dictionary<string, string>();
             this.IsPending = new CalendarEventStatuses().PendingForType(type).Contains(status);
         }
     }
--- a/server/Arcadia.Assistant.Calendar.Notifications/DictionaryExtensions.cs
+++ b/server/Arcadia.Assistant.Calendar.Notifications/DictionaryExtensions.cs
@@ -3,17 +3,16 @@
     using System.Collections.Generic;
     using System.Linq;
 
-    public static class DictionaryExtensions
+    public class DictionaryMerge
     {
-        public static IDictionary<T1, T2> Merge<T1, T2>(
-            this IDictionary<T1, T2> target,
+        public Dictionary<T1, T2> Perform<T1, T2>(
+            IEnumerable<KeyValuePair<T1, T2>> target,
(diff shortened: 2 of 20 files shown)


server/Arcadia.Assistant.Calendar.Abstractions/CalendarEvent.cs
```diff
@@ -16,18 +17,22 @@ public class CalendarEvent
 
         public string EmployeeId { get; }
 
+        public IReadOnlyDictionary<string, string> AdditionalData { get; }
```
santee: i'd have it set to non-null value by default, so we avoid all stupid NRE

server/Arcadia.Assistant.Calendar.Notifications/DictionaryExtensions.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace Arcadia.Assistant.Calendar.Notifications
+{
+    using System.Collections.Generic;
+    using System.Linq;
+
+    public static class DictionaryExtensions
+    {
+        public static IDictionary<T1, T2> Merge<T1, T2>(
```
santee: wtf? `target.Concat(source).GroupBy(x => x.key).ToDictionary(d => d.Key, d => d.Last().Value);`

santee: you realize extension methods are evil? `class DictionaryMerge(IDictionary a, IDictionary b).Perfrom()`

server/Arcadia.Assistant.Calendar.Notifications/TemplateExpressionParser.cs
```diff
@@ -0,0 +1,49 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text.RegularExpressions;
+
+    public static class TemplateExpressionParser
+    {
+        private static readonly Regex ifBlockRegex = new Regex("(?<startIfGroup>{if:(?<startIfToken>.*?)}).*?(?<endIfGroup>{endif:(?<endIfToken>.*?)})");
+
+        public static string ParseTemplateExpression(this string template, IDictionary<string, string> context)
```
santee: so you apply it to ANY string.

sponkrashin: And what do you suggest? Use it like `new TemplateExpressionParser().Parse(...)`?

server/Arcadia.Assistant.Calendar.Notifications/DictionaryExtensions.cs
sponkrashin: Why are the evil? If I know that this method can be applied to any dictionaries and arrays, why not?

server/Arcadia.Assistant.Calendar.Notifications/TemplateExpressionParser.cs
santee: yes

sponkrashin: Ugly.

server/Arcadia.Assistant.Calendar.Notifications/DictionaryExtensions.cs
santee: they cannot be DI'ed (you probably don't need to anyway :) ).
Merge here is not the worst example though. Your parser is.
You don't write static class with static methods even if you know that they won't contain any static fields, right?

santee: the proper way to use extensions is fluent interfaces :)

server/Arcadia.Assistant.Calendar.Notifications/TemplateExpressionParser.cs
santee: no. properly.

server/Arcadia.Assistant.Calendar.Notifications/DictionaryExtensions.cs
santee: Also, all dictionary methods change the object itself. this one does not and it might confuse.

sponkrashin: But what do you have against
> static class with static methods even if you know that they won't contain any static fields

?

server/Arcadia.Assistant.Calendar.Abstractions/CalendarEvent.cs
sponkrashin: Done.

server/Arcadia.Assistant.Calendar.Notifications/DictionaryExtensions.cs
sponkrashin: Done.

server/Arcadia.Assistant.Calendar.Notifications/TemplateExpressionParser.cs
sponkrashin: Done.

PR #577, opened 2019-04-04
Feature/#559 make employee name clickable as well as photo

YaroTs:


mobile/src/feeds/feed-message.tsx
```diff
@@ -23,26 +23,18 @@ export class FeedMessage extends React.Component<FeedMessageProps> {
         const photo = employee ? employee.photoUrl : undefined;
         const formattedDate = message.datePosted.format('MMMM D, YYYY');
 
-        const isDisabledClick = !this.props.employee;
```
antonm76: Do you think this condition check is not needed anymore?

YaroTs: This condition is also being checked in the onAvatarClicked function provided to TouchableOpacity

Issue #562, opened 2019-03-28
Increase top margin on the tab bar


DesmanLead:

KuchaLeni:
Fixed

PR #584, opened 2019-04-05
#562 Increase top margin on the tab bar

antonm76:


antonm76:
Tab bar icons size is decrease on iOS and increased on Android:
Screenshot 2019-04-05 at 14 06 12
Screenshot 2019-04-05 at 14 06 21
Screenshot 2019-04-05 at 14 06 27
android-MainActivity-04052019141112

Issue #557, opened 2019-03-28
Tab bar: make currently selected tab icon more visible


DesmanLead:

YaroTs:
We already use bold icons for selected tab, but we can improve visibility by making label font bold and increasing text size 

Was:

![Screenshot 2019-04-05 at 14 29 17](https://user-images.githubusercontent.com/5641071/55624633-3e710b80-57af-11e9-8dbc-e9fb734f4806.png)

With bold font: 

![Screenshot 2019-04-05 at 14 17 35](https://user-images.githubusercontent.com/5641071/55624706-69f3f600-57af-11e9-89dc-c77006bed20a.png)

@antonm76 @DesmanLead @santee @KuchaLeni Any objections/thoughts?

YaroTs:
Well, as the proposed changes anyway had been already made during investigations, I've created a branch https://github.com/SoftwareCountry/arcadian-assistant/tree/feature/%23557_Tab_bar_make_currently_selected_tab_more_visible with them. Once we decide whether or not it's enough to change only tab text style, we are free to merge this branch or delete it otherwise.

KuchaLeni:
Fixed

PR #614, opened 2019-04-16
#557 Bold text style and increased text size applied for selected tab

antonm76:


antonm76:
Simulator Screen Shot - iPhone 5s - 2019-04-16 at 13 19 33

PR #616, opened 2019-04-17
Bugfix/609 completed button sick leave permissions driven

victor-kotov-arcadia:
Only users with appropriate permissions could Complete Sick leave.

### DIFF
--- a/mobile/src/calendar/event-dialog/edit-sick-leave-event-dialog.tsx
+++ b/mobile/src/calendar/event-dialog/edit-sick-leave-event-dialog.tsx
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import { EventDialogBase, eventDialogTextDateFormat } from './event-dialog-base';
-import { AppState, getEmployee, getPermission } from '../../reducers/app.reducer';
+import { AppState, getEmployee, hasPermission } from '../../reducers/app.reducer';
 import { Action, Dispatch } from 'redux';
 import { connect } from 'react-redux';
 import { closeEventDialog, openEventDialog } from '../../reducers/calendar/event-dialog/event-dialog.action';
@@ -113,7 +113,7 @@ const mapStateToProps = (state: AppState): EditSickLeaveEventDialogProps => {
     return {
         intervals: state.calendar ? state.calendar.calendarEvents.selectedIntervalsBySingleDaySelection : undefined,
         userEmployee: getEmployee(state),
-        completeSickLeavePermission: getPermission(state, Permission.completeSickLeave)
+        completeSickLeavePermission: hasPermission(Permission.completeSickLeave, state)
     };
 };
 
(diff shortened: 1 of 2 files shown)


mobile/src/calendar/event-dialog/edit-sick-leave-event-dialog.tsx
```diff
@@ -31,14 +33,18 @@ class EditSickLeaveEventDialogImpl extends Component<EditSickLeaveEventDialogPro
     public render() {
         const selectedSickLeave = this.selectedSickLeave();
         const isEditable = selectedSickLeave ? !selectedSickLeave.calendarEvent.isCompleted : false;
+        let isDisabledForComplete = true;
+        if (this.props.completeSickLeavePermission === true && !isEditable) {
+            isDisabledForComplete = false;
+        }
```
YaroTs: This code can be simplified with ternary operator or even just to something like
```
const isDisabledForComplete = !(this.props.completeSickLeavePermission === true && !isEditable)
```

DesmanLead: @YaroTs IMO, it doesn't look simple %)

mobile/src/calendar/event-dialog/edit-sick-leave-event-dialog.tsx
```diff
@@ -31,14 +33,18 @@ class EditSickLeaveEventDialogImpl extends Component<EditSickLeaveEventDialogPro
     public render() {
         const selectedSickLeave = this.selectedSickLeave();
         const isEditable = selectedSickLeave ? !selectedSickLeave.calendarEvent.isCompleted : false;
+        let isDisabledForComplete = true;
```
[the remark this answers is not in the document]
victor-kotov-arcadia: It's about button state not the sick leave - so it could be completed but only by user with proper permission, not for all. So IMO "canBeCompleted" is also could be confusing.

mobile/src/calendar/event-dialog/edit-sick-leave-event-dialog.tsx
```diff
@@ -31,14 +33,18 @@ class EditSickLeaveEventDialogImpl extends Component<EditSickLeaveEventDialogPro
...
+        if (this.props.completeSickLeavePermission === true && !isEditable) {
```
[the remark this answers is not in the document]
victor-kotov-arcadia: Addressed

mobile/src/calendar/event-dialog/edit-sick-leave-event-dialog.tsx
```diff
@@ -30,15 +32,16 @@ class EditSickLeaveEventDialogImpl extends Component<EditSickLeaveEventDialogPro
     //----------------------------------------------------------------------------
     public render() {
         const selectedSickLeave = this.selectedSickLeave();
-        const isEditable = selectedSickLeave ? !selectedSickLeave.calendarEvent.isCompleted : false;
+        const isCompleted = selectedSickLeave ? !selectedSickLeave.calendarEvent.isCompleted : false;
+        const isDisabledForComplete = !(this.props.completeSickLeavePermission === true && !isCompleted);
```
antonm76: IMO, it is better to avoid inverse logic wherever it is possible. Doesn't you think that
`const isDisabledForComplete = this.props.completeSickLeavePermission === false || isCompleted;`
is much more readable?

PR #622, opened 2019-04-18
Feature/#619 pin code system tab is white on i os

YaroTs:


mobile/src/splash-screen/splash-screen.tsx
```diff
@@ -91,6 +91,17 @@ class SplashScreenImpl extends React.Component<SplashScreenStateProps & SplashSc
 
     //----------------------------------------------------------------------------
     public render() {
+        const splashContent = this.renderContent();
```
antonm76: Why do you need `splashContent` variable here? Isn't it possible to call `this.renderContent()` inline?

DesmanLead:
Now there is a blue stripe at the top, right?
Did you consider setting the status bar text color to black?

YaroTs:
Yes, it is. Why the text color should be black?

antonm76:
Just make two screenshots and ask @KuchaLeni what is better.

DesmanLead:
I mean, don't add a blue stripe. Just black text on the white background.

mobile/src/splash-screen/splash-screen.tsx
YaroTs: Not sure how to do it correctly. Such code
```
<SafeAreaView style={splashScreenStyles.container}>
                {this.renderContent()}
            </SafeAreaView>
```
produce Top-level 'this' expression warning

Issue #638, opened 2019-04-23
Add sick leave prolong accounting notification


sponkrashin:
When sick leave is prolonged by employee, notification email should be sent to accounting department.

santee:
@sponkrashin add description :)

santee:
accounting dep should be notified about sick leaves prolongation, I guess

sponkrashin:
Right.

KuchaLeni:
Fixed
![image](https://user-images.githubusercontent.com/28759011/60025666-c70e9200-96a2-11e9-9703-4448d418c060.png)

Issue #637, opened 2019-04-23
Add sick leave end date to accounting notification


sponkrashin:

KuchaLeni:
![image](https://user-images.githubusercontent.com/28759011/60025803-105ee180-96a3-11e9-9632-a39dda14c08f.png)
![image](https://user-images.githubusercontent.com/28759011/60025816-181e8600-96a3-11e9-8dc2-5cebaad62ab1.png)

Issue #597, opened 2019-04-09
Approved sick leave notification: add additional employee identification in email


KuchaLeni:
Add additional employee identification in accounting email about approved sick leave, for ex. Employee ID (This is necessary to distinguish between employees with the same name and surname).

sponkrashin:
@santee, any thoughts?

santee:
employee id

KuchaLeni:
Fixed
![image](https://user-images.githubusercontent.com/28759011/60026213-d8a46980-96a3-11e9-851f-4b3fba6db99d.png)

PR #639, opened 2019-04-23
Feature/sick leave notifications

sponkrashin:
Issues #637, #638, #597


server/Arcadia.Assistant.Calendar.Notifications/SickLeaveAccountingEmailNotificationActor.cs
```diff
@@ -0,0 +1,121 @@
+﻿namespace Arcadia.Assistant.Calendar.Notifications
```
sponkrashin: It is not new, it is just renamed.

PR #662, opened 2019-05-07
When searching employee by identity used both checks

sponkrashin:
One for modern identity (email field) and one for legacy identity (login field + domain).

### DIFF
--- a/server/Arcadia.Assistant.Organization.Abstractions/EmployeeMetadata.cs
+++ b/server/Arcadia.Assistant.Organization.Abstractions/EmployeeMetadata.cs
@@ -1,15 +1,15 @@
 ﻿namespace Arcadia.Assistant.Organization.Abstractions
 {
     using System;
+    using System.Collections.Generic;
     using System.Diagnostics;
 
     public class EmployeeMetadata
     {
-        public EmployeeMetadata(string employeeId, string name,  string legacyIdentity, string email)
+        public EmployeeMetadata(string employeeId, string name, string email)
         {
             this.EmployeeId = employeeId;
             this.Name = name;
-            this.LegacyIdentity = legacyIdentity;
             this.Email = email;
         }
 
@@ -17,9 +17,9 @@
 
         public string Name { get; set; }
 
-        public string LegacyIdentity { get; }
-
         public string Email { get; set; }
+
+        public IEnumerable<string> IdentityAliases { get; set; }
 
         public string RoomNumber { get; set; }
 
--- a/server/Arcadia.Assistant.Organization.Abstractions/EmployeeMetadata.cs
+++ b/server/Arcadia.Assistant.Organization.Abstractions/EmployeeMetadata.cs
@@ -5,11 +5,11 @@
 
     public class EmployeeMetadata
     {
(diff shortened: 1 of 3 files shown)


server/Arcadia.Assistant.Organization.Abstractions/EmployeeMetadata.cs
```diff
@@ -1,26 +1,26 @@
...
 
         public string EmployeeId { get; }
 
         public string Name { get; set; }
 
-        public string Identity { get; set; }
-
         public string Email { get; set; }
 
+        public IEnumerable<string> IdentityAliases { get; set; }
```
santee: default value maybe?

server/Arcadia.Assistant.Organization/EmployeeSearch.cs
```diff
@@ -110,9 +110,13 @@ private static bool CheckFilter(EmployeeMetadata employee, EmployeesQuery query)
                 return false;
             }
 
-            if ((query.Identity != null) && !string.Equals(employee.Identity, query.Identity, StringComparison.InvariantCultureIgnoreCase))
+            if (query.Identity != null)
             {
-                return false;
+                if (!string.Equals(employee.Email, query.Identity, StringComparison.InvariantCultureIgnoreCase) &&
+                    employee.IdentityAliases?.Any(x => string.Equals(x, query.Identity, StringComparison.InvariantCultureIgnoreCase)) != true)
```
santee: i suggest hashset with ignorecase comparer. when its just Contains without any lambdas.

server/Arcadia.Assistant.Organization.Abstractions/EmployeeMetadata.cs
sponkrashin: I made such at first, but anyway it can be set to null and we need null-propagation, so why do we need that default value?

server/Arcadia.Assistant.Organization/EmployeeSearch.cs
sponkrashin: Always forget about those cool collections :)

server/Arcadia.Assistant.Organization.Abstractions/EmployeeMetadata.cs
santee: i would not use null propagation, that's all :)

santee: basically, I'd try to avoid nulls if possible. '?' makes it safe but uncertain

sponkrashin: Everyone needs not-null reference values in C# then :)

server/Arcadia.Assistant.Organization/EmployeeSearch.cs
sponkrashin: Done.

server/Arcadia.Assistant.Organization.Abstractions/EmployeeMetadata.cs
sponkrashin: Done.

PR #685, opened 2019-05-29
User profile: events list - add day of the week to date string

victor-kotov-arcadia:
short day name ahead of date at event details


DesmanLead:
Is the same pattern used on profile screen? As far as I remember there are separate screens: employee details and profile.

victor-kotov-arcadia:
> Is the same pattern used on profile screen? As far as I remember there are separate screens: employee details and profile.

Yes, that was my assumption too, but from my tests it seems to be fine now - I've checked my profile and also checked it via Company->My details - both were showing it with week day.

victor-kotov-arcadia:
> Don't we want such behaviour also in dialogs like "Your sick leave start on ..."?

Probably yes, but it wasn't mentioned in initial request

Issue #677, opened 2019-05-29
Calendar: mark weekends


DesmanLead:
Mark weekends (day numbers and names) with a gray color (like iOS calendar).

KuchaLeni:
fixed

KuchaLeni:
![image](https://user-images.githubusercontent.com/28759011/60025354-1f915f80-96a2-11e9-83c0-396a1fdf1ad1.png)

Issue #676, opened 2019-05-29
Calendar: mark today


DesmanLead:
Probably making it bold will be fine.

KuchaLeni:
Fixed.
![image](https://user-images.githubusercontent.com/28759011/60025433-49e31d00-96a2-11e9-974c-b3bc5cc0fef2.png)
![image](https://user-images.githubusercontent.com/28759011/60025446-4fd8fe00-96a2-11e9-8a9a-275e0a520990.png)

Issue #675, opened 2019-05-29
Calendar: start week from Monday


DesmanLead:
Or get system settings (Mon or Sun) if it is not too complicated.

KuchaLeni:
Fixed

PR #687, opened 2019-05-30
Feature/calendar screen

YaroTs:
#675 Calendar: start week from Monday
#676 Calendar: mark today
#677 Calendar: mark weekends

Week starts from Sunday only in case of USA English locale on device.

Before: 

Screenshot 2019-05-30 at 13 58 07

After: 

Screenshot 2019-05-30 at 18 21 14


mobile/src/calendar/calendar-page-weekday.tsx
```diff
@@ -85,11 +87,31 @@ export class WeekDayCircle extends Component<WeekDayCircleProps> {
...
+
+            if (this.isSelectedDay(day)) {
+                return Style.color.white;
+            }
+
+            if (business.isWeekendDay(day.date)) {
+                return Style.color.gray;
+            }
+
+            if (customTextColor) {
```
antonm76: Why check for custom text color is on the last place? Shouldn't custom color win?

mobile/src/locale/locale.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Copyright (c) Arcadia, Inc. All rights reserved.
+ ******************************************************************************/
+
+import { NativeModules, Platform } from 'react-native';
+import * as moment from 'moment';
+import 'moment/locale/en-gb';
+
+//----------------------------------------------------------------------------
+export function applyLocale() {
+    const defaultLocale = 'en-gb';
```
antonm76: Why 'en-gb'? Just to have some default value? How will it affect UI?

DesmanLead: A week starts from Monday in GB :)

antonm76: Ah, I see.

mobile/src/calendar/calendar-page-weekday.tsx
YaroTs: This variable really has confusing name and actually wasn't used. I've removed in e560f680693caf5e9822d20e26502b19a80f0581

Issue #569, opened 2019-04-01
'Room null' is displayed for employees without rooms


KuchaLeni:
for ex. employee from Minsk
![image](https://user-images.githubusercontent.com/28759011/55329313-f5f7dc00-5496-11e9-8b7f-c6fcb21ed57c.png)

DesmanLead:
@KuchaLeni @santee what should be displayed there? %)

santee:
let's put a question mark :D 'Room <?>'

PR #688, opened 2019-05-31
#569 'Room null' is displayed for employees without rooms

YaroTs:


mobile/src/reducers/organization/employee.model.ts
```diff
@@ -99,16 +99,26 @@ export class Employee {
...
 
+    //----------------------------------------------------------------------------
     public getName(): string {
         const names = this.name.split(' ');
 
         return names.length > 1 ? names[1] : this.name;
     }
 
+    //----------------------------------------------------------------------------
+    public getRoomTitle(): string {
```
DesmanLead: I'm a bit concerned that there are 2 places with room number string generation. Maybe it's better to extract this somewhere?

YaroTs: Sounds reasonable, but where?

DesmanLead: That's a good question. Some kind of RoomFormatter?

antonm76: Maybe pass fully constructed room title into CurrentPeopleRoomImpl as a nav param would be a solution? I see CurrentPeopleRoomImpl is the only place where room title is constructed besides this function

YaroTs: You should still provide some default value if there is no required parameter in payload in this case

YaroTs: @DesmanLead Take a look at latest changes, please

Issue #691, opened 2019-05-31
Profile: add events status description


DesmanLead:
**Vacation**
*  "Requested" waiting for approvals
*  "Approved" preparing documents
*  "Cancelled" cancelled
*  "Rejected" rejected
*  "AccountingReady" documents are ready to sign
*  "Processed" confirmed

**Dayoff / Workout**
*  "Requested" waiting for approvals
*  "Approved" confirmed
*  "Cancelled" cancelled
*  "Rejected" rejected

**Sickleave**
*  "Requested" confirmed
*  "Completed" completed
*  "Cancelled" cancelled

KuchaLeni:
Fixed

Issue #694, opened 2019-06-03
Change vacations approval process


santee:
Change vacations approval as follows:
1. Only one approval is required for request to be approved
2. Upon creation it goes to direct manager's pending actions list
3. If request is not approved in configured amount of working days (say, 3 days by default), it goes to the next approver's pending actions list (but stays in manager's list though).

sponkrashin:
Approvals work the following way:
All event types need one approval of any employee that have rights to do it.
Vacations have specific flow - if assigned approver won't approve event after some days (3 by default settings), event is assigned to next approver from departments hierarchy (for my event approvers are Shevnin, Maxim, Efimova, Baklan), but it still is shown in previous approver pending actions list.

KuchaLeni:
Fixed

Issue #679, opened 2019-05-29
Remove approval step from sick-leaves


santee:
Following changes need to be implemented:
1. No approval is required in sick-leave (mobile should display it as approved immediately). @DesmanLead see if you want to create a separate task to track it
2. Email to HR should be sent upon sick leave submission
3. Notification (push/email) should be sent to a manager upon submission
4. Emails to HR (and manager) should be sent whenever it's prolonged or cancelled

sponkrashin:
1. Will probably be done automatically, because server will return event with status Approved on create request like it works for all Baklan requests for now.

KuchaLeni:
Related issues: #712

KuchaLeni:
Fixed

PR #695, opened 2019-06-03
#691 Profile: add events status description

YaroTs:
#691 Profile: add events status description

Screenshot 2019-06-03 at 16 00 40

### DIFF
--- a/mobile/src/employee-details/employee-details-events-list.tsx
+++ b/mobile/src/employee-details/employee-details-events-list.tsx
@@ -40,29 +40,26 @@ class EmployeeDetailsEventsListImpl extends Component<EmployeeDetailsEventsListP
         [CalendarEventStatus.Approved]: 'preparing documents',
         [CalendarEventStatus.AccountingReady]: 'documents are ready to sign',
         [CalendarEventStatus.Completed]: 'completed',
-        [CalendarEventStatus.Processed]: 'all set',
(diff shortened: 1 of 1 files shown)


mobile/src/employee-details/employee-details-events-list.tsx
```diff
@@ -29,8 +32,39 @@ interface EmployeeDetailsEventsListDispatchProps {
 
 //============================================================================
 class EmployeeDetailsEventsListImpl extends Component<EmployeeDetailsEventsListProps & EmployeeDetailsEventsListDispatchProps> {
+
     private readonly eventDigitsDateFormat = 'ddd DD/MM/YYYY';
 
+    private readonly vacationDescriptions: CalendarEventStatusDescriptions = {
+        [CalendarEventStatus.Requested]: 'waiting for approvals',
```
santee: is it possible to split CalendarEventStatus on three different types? Changes are coming from server side when you'll have to change this mapping

YaroTs: What kind of server changes do you mean?

santee: #694 #679 
Even bigger difference in statuses set.

antonm76: It would be great to just have a table with all statuses and their flows for particular events types instead of digging other PRs. -)

Issue #701, opened 2019-06-06
Update sick leave status handling (approval step was removed)


DesmanLead:
Mobile counterpart for #679.

Possible sick leave statuses now are:
[
  "Requested",
  "Completed",
  "Cancelled"
]

Requested should be treated as approved in UI.

KuchaLeni:
Related issue #679

KuchaLeni:
Fixed

PR #702, opened 2019-06-06
Feature/calendar event status

YaroTs:
#701 Update sick leave status handling (approval step was removed)


mobile/src/employee-details/employee-details-events-list.tsx
```diff
@@ -198,11 +194,11 @@ class EmployeeDetailsEventsListImpl extends Component<EmployeeDetailsEventsListP
         }
 
         switch (event.status) {
-            case CalendarEventStatus.Requested:
-            case CalendarEventStatus.Approved:
-                return CalendarEventStatus.Requested;
-            case CalendarEventStatus.Processed:
-                return CalendarEventStatus.Approved;
```
DesmanLead: It was there before so I don't think it's a good idea to change this in the same PR. But this code is, well... surprising.

Issue #671, opened 2019-05-29
Handle new release notifications


DesmanLead:
Mobile part of #656

sponkrashin:
After new build is downloaded by server , push notification will be sent with custom data parameter `Type = "UpdateAvailable"`.

DesmanLead:
@sponkrashin `Type`, not `type`? All other fields look like: `employeeId`, `approverId`.

sponkrashin:
I send it as "Type", but serializer, probably, makes "type" from it. Anyway, all is the same as for other notifications.

antonm76:
@YaroTs the full list of notification types can be found here: https://github.com/SoftwareCountry/arcadian-assistant/blob/master/server/Arcadia.Assistant.Calendar.Notifications/CalendarEventPushNotificationTypes.cs

KuchaLeni:
Android:
![image](https://user-images.githubusercontent.com/28759011/60724848-26409200-9f40-11e9-86b7-e1df110a5c21.png)

iOS:
![image](https://user-images.githubusercontent.com/28759011/60725173-0f4e6f80-9f41-11e9-9e0a-e06c47e3365e.png)

PR #706, opened 2019-06-10
#671 Handle new release notifications

YaroTs:


mobile/src/notifications/notifications.ts
```diff
@@ -0,0 +1,15 @@
+/******************************************************************************
+ * Copyright (c) Arcadia, Inc. All rights reserved.
+ ******************************************************************************/
+
+//============================================================================
+export enum NotificationType {
+    UpdateAvailable = 'UpdateAvailable',
```
antonm76: Consider enum variants to be started from a lowercase letter

mobile/src/notifications/notification.epics.ts
```diff
@@ -93,11 +104,11 @@ const notificationsRegister$ = (action$: ActionsObservable<NotificationAction>,
             android: 'Android',
         });
         return deps.apiClient.post(`/push/device`,
-                                   {
-                                       devicePushToken: action.installId,
-                                       deviceType,
-                                   },
-                                   { 'Content-Type': 'application/json' }).pipe(
+            {
```
antonm76: Current tslint settings do not allow such formatting :-(((

mobile/src/notifications/notification.epics.ts
```diff
@@ -28,40 +29,50 @@ const notificationsHandler$ = (action$: ActionsObservable<LoadUserFinished>, sta
...
-                                observer.next(openProfile());
-                            }
-                        } else {
-                            logError('Unexpected notification payload', notification.customProperties);
+                        switch (type) {
+                            case NotificationType.UpdateAvailable:
+                                if (!url) {
+                                    logError('Unexpected notification payload', notification.customProperties);
+                                } else {
+                                    Linking.openURL(url);
```
antonm76: Add // noinspection JSIgnoredPromiseFromCall above

santee: add a catch instead? So it doesn't crash for weird reason? `Linking.openUrl(url).catch( console.error )`

YaroTs: It can crash only in case of none-http links, but anyway it's really better to handle it instead of ignoring. c6ada5463594568dcb2f3b2ce5e75d4087bb7766

DesmanLead:
@sponkrashin check out NotificationType enum, please. Does it correspond to current server implementation?

PR #713, opened 2019-06-27
Sick leave: 'Accept' and 'Decline' buttons are displayed in employee profile

DesmanLead:


mobile/src/employee-details/event-action-provider.ts
```diff
@@ -91,6 +91,13 @@ export class EventActionProvider {
                            permissions: UserEmployeePermissions,
                            approvals: Map<CalendarEventId, Set<Approval>>): Optional<EventAction> {
 
+        // Only the event creator should be able to cancel the sick leave. See:
+        // https://github.com/SoftwareCountry/arcadian-assistant/issues/712
+        const isOwnEvent = this.isCurrentUser(employee);
+        if (event.isSickLeave && !isOwnEvent) {
```
sponkrashin: If I understood correctly, this method should return undefined for each request if type is sick leave. Sick leave cannot be approved at all only cancelled.

DesmanLead: @sponkrashin Looking at the questions this code raises I think it's better to add a separate branch for sick leaves there. When sick leave can be cancelled? Only when it isOwnEvent and status != completed?

PR #766, opened 2020-01-16
Upgrade to React Native 61.5

DesmanLead:


YaroTs (approved):
You definitely should wait for dev server to work normally to check all this changes before merging)

